PR #8, opened 2021-12-10
Add dra, ddec, and distance to PrecoveryCandidate

### LINKED ISSUE #5: Add dra, ddec, and distance to PrecoveryCandidate
Additional useful quantities to include in the [PrecoveryCandidate](https://github.com/B612-Asteroid-Institute/precovery/blob/master/precovery/precovery_db.py#L18-L25) dataclass are the sky-plane residuals and the great circle distance.

moeyensj:
Work to address #5. This PR adds dra, ddec, and distance.

### DIFF
--- a/precovery/frame_db.py
+++ b/precovery/frame_db.py
@@ -176,6 +176,11 @@ class FrameIndex:
     ) -> Iterator[HealpixFrame]:
         """
         Yield all the frames which are for given obscode, MJD, healpix.
+
+        MJDs are checked to within +- 1e-7 days or 8.64 ms. Any frames that
+        are within 8.64 ms of the given mjd will be returned. This does not garauntee
+        that they will represent the desired exposure time and may lead to multiple
+        matches computed at the wrong observation time.
         """
         select_stmt = sq.select(
             self.frames.c.id,
@@ -189,8 +194,8 @@ class FrameIndex:
         ).where(
             self.frames.c.obscode == obscode,
             self.frames.c.healpixel == int(healpixel),
-            self.frames.c.mjd >= mjd - 0.001,
-            self.frames.c.mjd <= mjd + 0.001,
+            self.frames.c.mjd >= mjd - 1e-7,
+            self.frames.c.mjd <= mjd + 1e-7,
         )
         rows = self.dbconn.execute(select_stmt)
         for r in rows:
--- a/precovery/frame_db.py
+++ b/precovery/frame_db.py
@@ -71,8 +71,10 @@ class Observation:
         """
         return cls(ra=so.ra, dec=so.dec, id=so.id)
 
-    def matches(self, ephem: Ephemeris, tolerance: float) -> bool:
+    def matches(self, ephem: Ephemeris, tolerance: float) -> Tuple[bool, float, float, float]:
         distance = haversine_distance_deg(self.ra, ephem.ra, self.dec, ephem.dec)
+        dra = ephem.ra - self.ra
+        ddec = ephem.dec - self.dec
         logger.debug(
             "%.4f, %.4f -> %.4f, %.4f = %.6f\t(tol=%.6f)",
             self.ra,
@@ -82,7 +84,7 @@ class Observation:
             distance,
             tolerance,
         )
-        return distance < tolerance
+        return distance < tolerance, dra, ddec, distance
 
 
 class FrameIndex:
--- a/precovery/frame_db.py
+++ b/precovery/frame_db.py
@@ -204,7 +204,21 @@ class FrameIndex:
             self.frames.c.mjd >= mjd - 1e-7,
             self.frames.c.mjd <= mjd + 1e-7,
         )
-        rows = self.dbconn.execute(select_stmt)
+        result = self.dbconn.execute(select_stmt)
+        # Turn result into a list so we can iterate over it twice: once
+        # to check the MJDs for uniqueness and a second time to actually
+        # yield the individual rows
+        rows = list(result)
+
+        # Loop through rows and track MJDs
+        mjds = set()
+        for r in rows:
+            # id, obscode, catalog_id, mjd, healpixel, data uri, data offset, data length
+            mjds.add(r[3])
+
+        if len(mjds) > 1:
+            logger.warn(f"Query returned non-unique MJDs for mjd: {mjd}, healpix: {int(healpixel)}, obscode: {obscode}.")
+
         for r in rows:
             yield HealpixFrame(*r)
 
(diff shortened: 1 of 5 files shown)


precovery/precovery_db.py
```diff
@@ -207,14 +210,18 @@ def _check_frames(
             n = 0
             for obs in self.frames.iterate_observations(f):
                 n += 1
-                if obs.matches(exact_ephem, tolerance):
+                matches, dra, ddec, distance = obs.matches(exact_ephem, tolerance)
```
moeyensj: Would it be cleaner if we redefine `obs.matches` to `obs.distance` and have this function return `dra, ddec, distance`? We can then delegate the check against `tolerance` in this for-loop rather than returning the boolean.

moeyensj:
@spenczar Could you take a look at the changes to the `obs.matches` function and let me know your thoughts? There are two other commits in this PR, one that fixes a timedelta bug and another that disables the linear propagator. Disabling the linear propagator is a temporary change while we push forward on getting some of the candidates longer arcs of observation. Falling back to 2-body as our "approximate" propagation has made it so this code agrees 1-1 with the observations from brute-force attribution.

precovery/precovery_db.py
spenczar: Yes, definitely. `matches` should return just a `bool`, based on the name. `distance` seems like a good name for a method that computes distances.

precovery/frame_db.py
```diff
@@ -174,6 +176,11 @@ def get_frames(
     ) -> Iterator[HealpixFrame]:
         """
         Yield all the frames which are for given obscode, MJD, healpix.
+
+        MJDs are checked to within +- 1e-7 days or 8.64 ms. Any frames that
+        are within 8.64 ms of the given mjd will be returned. This does not garauntee
+        that they will represent the desired exposure time and may lead to multiple
+        matches computed at the wrong observation time.
```
spenczar, line 190: Oof, this is annoying, isn't it?

This is just a problem with floating point timestamps, I suppose. Perhaps there is a way we can log at the WARN level if we get multiple hits?

precovery/orbit.py
```diff
@@ -249,7 +249,7 @@ def approximately_propagate(
 
         Accuracy will decrease as timedelta increases.
         """
-        if timedelta <= 3.0 and self.ra_velocity < 1 and self.dec_velocity < 1:
+        if np.abs(timedelta) <= -1.0 and self.ra_velocity < 1 and self.dec_velocity < 1:
```
spenczar, line 252: Good catch to use `np.abs` - but why the change of the RHS to `<= -1`? It seems like that expression will always be `false` now.

spenczar (changes requested):
Some good catches here, but I think moving the distance calculations out of `matches` and into a new `distance` method seems much better.

precovery/orbit.py
moeyensj: This is just a temporary change to disable the approximate propagator. Disabling the approximate propagation helped recover missing observations for some of the objects we were looking at. The more sophisticated approach would be to analyze the trade-off between healpixel size, approximate propagation time scale, and the predicted RA and Dec velocites. For the time being, however, using 2-body by default and then N-body propagations for the more precise propagations is thankfully fast enough.

precovery/precovery_db.py
moeyensj: Done, I've changed the matches function to distance and move the check against tolerance to the for-loop in [f4ad4d5](https://github.com/B612-Asteroid-Institute/precovery/pull/8/commits/f4ad4d53b5363eefa9b51ba681a4795edb5f481b).

precovery/frame_db.py
moeyensj: I'd suggest we reduce the 1e-7 further to 1e-12 or 0.0864 microseconds (what is the smallest comparison we could go to numerically?). We should hopefully never (🤞) encounter two exposures within 0.0864 microseconds from the same observatory. 

Checking for MJD uniqueness would require (I think) looping over the rows first before yielding them back and tracking the unique number of MJDs. It's not fully obvious how much more expensive that would be but it is something that is totally doable. Something as simple as this could work? 
```
rows = self.dbconn.execute(select_stmt)

mjds = set()
for r in rows:
    mjds.add(r[3])

if len(mjds) > 1:
   logger.warn(f"Query returned non-unique MJDs for mjd: {mjd}, healpix: {int(healpixel)}, obscode: {obscode}.")
 
for r in rows:
    yield HealpixFrame(*r)
```

precovery/precovery_db.py
moeyensj: Done, I've changed the matches function to distance and moved the check against tolerance to the for-loop in [f4ad4d5](https://github.com/B612-Asteroid-Institute/precovery/pull/8/commits/f4ad4d53b5363eefa9b51ba681a4795edb5f481b).

precovery/frame_db.py
spenczar: This is really fraught with peril. Floating point inaccuracies are really hard to predict. See https://www.sqlite.org/floatingpoint.html - a serious applied mathematician could probably put error bounds on this but that's way beyond my abilities.

Your loop is valid and fast, and the warning is probably sufficient. I think that's very preferable over shrinking down to less than a microsecond.

The real Right Thing here is to stop using floating point values. We should store times as integers if we want to do exact matching. That could be done by using something like UTC milliseconds, which should map cleanly from Julian Dates. Since sqlite uses 64-bit integers, this should be valid for essentially forever. Another option would be to multiply our MJDs by a constant integer factor (say, 1e9) and round to the nearest integer.

Then, we could actually do exact matches and thing should work correctly.

precovery/orbit.py
spenczar: Got it. Let's just remove approximate propagation entirely, then, and use 2-body. That will mean a lot of code deletion - which is a very good thing.

moeyensj: The code is fast enough for now but once we start indexing bigger datasets we may need to go back and look for faster methods. The linear propagation should work fine for time regimes that are small (hours to days, for most objects). Let's keep it for now so we have a starting place for future optimization.

precovery/frame_db.py
moeyensj: I've added a logging statement at the WARN level in [0799cd1](https://github.com/B612-Asteroid-Institute/precovery/pull/8/commits/0799cd128f1e8da670f7c003cb35294af3f8dcd2) and kept the time range at +- 1e-7 days.

moeyensj:
@spenczar I think this is ready for another look.

precovery/orbit.py
spenczar: Okay - but I still think this could be clearer. It could be (for example) `if False and np.abs(timedelta) <= 1.0 and ...`, or at least could have a comment. The current thing just looks very odd, and future optimization might come in, like, two years, and by different people, so we should be clearer here.

spenczar (approved):

moeyensj:
Closes #5.

PR #10, opened 2021-12-17
Add ra_sigma and dec_sigma to indexed observations

### LINKED ISSUE #9: Add ra_sigma, dec_sigma to observations
Adding uncertainties to the index observations will allow OD to be performed using `precovery` outputs rather than needing to return to the source observation files.

moeyensj:
This PR adds 1-sigma uncertainties to the indexed observations. I've tested this with the latest version of the observation HDF files.

Closes #9.

### DIFF
--- a/precovery/frame_db.py
+++ b/precovery/frame_db.py
@@ -14,8 +14,10 @@ from . import sourcecatalog
 from .orbit import Ephemeris
 from .spherical_geom import haversine_distance_deg
 
-logger = logging.getLogger("frame_db")
+# ra, dec, ra_sigma, dec_sigma, id
+DATA_LAYOUT = "<ddddl"
 
+logger = logging.getLogger("frame_db")
 
 @dataclasses.dataclass
 class HealpixFrame:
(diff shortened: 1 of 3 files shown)


spenczar (approved):
This looks good - glad it was so simple. 

A few further thoughts on this though:
 - This breaks the data file format; old data files will be invalid in strange ways (probably breaking because string IDs will seem completely invalid). Now is not the time for this, but at some point it will be worthwhile to store a bit of metadata which identifies the version being used so we can detect this. Maybe that's a new `metada.json` file, or maybe it belongs inside `index.db`.
 - There really should be tests that cover this behavior. I think [this test](https://github.com/B612-Asteroid-Institute/precovery/blob/master/tests/test_exposure.py#L13) at least will break with this change, but more tests would be good.

moeyensj:
> This breaks the data file format; old data files will be invalid in strange ways (probably breaking because string IDs will seem completely invalid). Now is not the time for this, but at some point it will be worthwhile to store a bit of metadata which identifies the version being used so we can detect this. Maybe that's a new metada.json file, or maybe it belongs inside index.db.

I'll file an issue so we remember to add this in the future. Tracking the version at indexing time is definitely something we are going to need to need as well when we index multiple datasets at the different times they may come available to us.

moeyensj:
> There really should be tests that cover this behavior. I think this test at least will break with this change, but more tests would be good. 

Couldn't agree more. I haven't touched any unit tests but we definitely need to update them (some of the tests are using objects and modules that no longer exist). This may be a good sprint or hack-session goal. I'll merge this PR and make a note to create a new unit test PR. The good thing is that it shouldn't be too difficult to get the coverage to 100% while the code is still in this stage of the development.

PR #15, opened 2022-03-01
Pyoorb init issue

KatKiker:


KatKiker:
May want to add a check to ensure the init was successful. 

I should be able to tackle that when I do the rest of the error handling.

PR #22, opened 2022-04-15
Predicted ephemeris and attribute naming improvements for PrecoveryCandidate, rename catalog_id to exposure_id, expose window_size

### LINKED ISSUE #7: Make window_size a search parameter
If the window size doesn't affect the underlying data structure of the precovery binary files and the index database, then exposing the window_size parameter to the precovery function would be useful to allow tuning while running precovery searches. 

Basically adding a keyword argument here: https://github.com/B612-Asteroid-Institute/precovery/blob/master/precovery/precovery_db.py#L78

### LINKED ISSUE #20: Add predicted ephemeris to PrecoveryCandidate
Returning the predicted RA and Dec and other useful observables as part of the PrecoveryCandidate dataclass will be useful for generating cutouts/postage stamps since those are queried on predicted positions.

moeyensj:
Addresses #20. 

This PR introduces a new schema for the PrecoveryCandidate dataclass that should make it a little more user-friendly:
```
class PrecoveryCandidate:
    mjd_utc: float
    ra_deg: float
    dec_deg: float
    ra_sigma_arcsec: float
    dec_sigma_arcsec: float
    mag: float
    mag_sigma: float
    filter: str
    obscode: str
    exposure_id: str
    observation_id: str
    healpix_id: int
    pred_ra_deg: float
    pred_dec_deg: float
    pred_vra_degpday: float
    pred_vdec_degpday: float
    delta_ra_arcsec: float
    delta_dec_arcsec: float
    distance_arcsec: float
```

The class now also carries the predicted location of the orbit, which will be useful for both generating cutouts and future negative observations analyses. 

Lastly, catalog_id has been renamed to exposure_id so that it is consistent throughout the code. Unfortunately, this means existing `index.db`s will need to be updated with a name change. Here is a Python snippet that will do that:
```
import sqlite3 as sql

INDEX_DB = ""

con = sql.connect(INDEX_DB)
con .execute("""ALTER TABLE frames RENAME COLUMN catalog_id TO exposure_id;""")
con.close()
```


moeyensj:
PR now also closes #7.

PR #25, opened 2022-04-21
Add FrameCandidate, add index_observations.py, calculate PrecoveryCandidate healpix_id at higher nside

### LINKED ISSUE #21: Add FrameCandidate
A potential use case is also returning "negative observations": instances where a given trajectory intersects a healpix frame but no observations were found. This only makes sense to do when the Healpixels are a small enough size but it could be worthwhile adding as an option to the user.

### LINKED ISSUE #24: Add higher nside HEALpixel IDs to PrecoveryCandidate
PrecoveryCandidate HEALpixel IDs are currently set to the HEALpixel ID of the HealpixFrame. If we decide to re-index the observations to a higher `nside` parameter then Healpixel IDs from previous versions will not match the new ones. The solution is to return Healpixel IDs with an `nside` parameter significantly higher than the `nside` parameters with which we would ever consider indexing the observations. At this point matching Healpixel IDs from a higher `nside` parameter to a lower `nside` parameter is a simple rshift. 

For example, 
```
import healpy as hp

k = 15
nside_15 = 2**k
nside_17 = 2**(k + 2)

id_15 = hp.ang2pix(nside_15, 0, 0, nest=True, lonlat=True)
id_17 = hp.ang2pix(nside_17, 0, 0, nest=True, lonlat=True)

print("id (k=15): ", id_15)
print("id (k=17): ", id_17)
print("id (k=17) >> 2 >> 2:", id_17 >> 2 >> 2)
```
Which returns:
```
id (k=15):  5100273664
id (k=17):  81604378624
id (k=17) >> 2 >> 2: 5100273664
```

moeyensj:
Addresses and closes #21 and #24.

### DIFF
--- a/precovery/__init__.py
+++ b/precovery/__init__.py
@@ -1,0 +1,1 @@
+from .main import precover
--- a/precovery/frame_db.py
+++ b/precovery/frame_db.py
@@ -107,31 +107,6 @@ class Observation:
             id=so.id,
         )
 
-    def distance(self, ephems: Iterable[Ephemeris]) -> Tuple[float, float, float]:
-        """
-        Calculate the Haversine distance and residuals in degrees between this observation
-        and a predicted ephemeris.
-        """
-        distance = haversine_distance_deg(
-            self.ra,
-            np.array([ephem.ra for ephem in ephems]),
-            self.dec,
-            np.array([ephem.dec for ephem in ephems]),
-        )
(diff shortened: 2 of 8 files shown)


moeyensj:
This PR has been tested on both my home system (with access to 3 months of data) and on epyc with the full dataset. Results are consistent and everything appears to be working.

PR #41, opened 2022-09-20
Create composite index on (mjd, healpixel, obscode)

### LINKED ISSUE #40: Fix (remove) single column indices and explicitly build composite index on (healpixel, mjd, obscode)
Timings on the newly indexed database (4 single indexes, 1 composite index):
```` 'ix_frames_obscode', 'ix_frames_mjd', ix_frames_healpixel', 'sqlite_autoindex_datasets_1', 'fast_query' ````
```
Found 29 potential matches for orbit ID 330784 in 139.495s
Found 33 potential matches for orbit ID 376552 in 136.716s
Found 22 potential matches for orbit ID 458508 in 137.527s
```

Timings on the new database (1 composite index):
```` 'fast_query' ````
```
Found 29 potential matches for orbit ID 330784 in 118.748s
Found 33 potential matches for orbit ID 376552 in 108.497s
Found 22 potential matches for orbit ID 458508 in 106.770s
```

Timings on the old database (1 composite index):
```` 'fast_query' ````
```
Found 29 potential matches for orbit ID 330784 in 111.993s
Found 33 potential matches for orbit ID 376552 in 109.673s
Found 22 potential matches for orbit ID 458508 in 107.731s
```
This [commit](https://github.com/B612-Asteroid-Institute/precovery/commit/afdfcab7023dedf28ee3e24e2acdc3dbdc27076e) did not have the intended effect of creating a faster query. We need to add a fix that creates a composite index equivalent to:
```
import sqlite3 as sql
con = sql.connect(DB)
con.execute("""CREATE INDEX fast_query ON frames (healpixel, mjd, obscode)""")
con.close()
```

### PR DESCRIPTION
**Pull request by moeyensj**
Also delete single indices and add index on dataset_id in datasets table. Closes #40. 

PR is currently being tested on the NSC dataset, will report back once complete.

### DIFF
--- a/precovery/frame_db.py
+++ b/precovery/frame_db.py
@@ -11,6 +11,7 @@ from typing import (
     Set,
     Tuple
 )
+import warnings
 
 import numpy as np
 import sqlalchemy as sq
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by moeyensj** [comment]
I tested this on a small sample of the NSC dataset (~4 hdf5 files indexed using this branch). 

```
import sqlite3 as sql

con = sql.connect("ndet4_test_00/index.db")
curs = con.execute("""SELECT name FROM sqlite_master WHERE type = 'index';""")
result = curs.fetchall()
print(curs.fetchall())
```
Which returns:
`[('fast_query',), ('sqlite_autoindex_datasets_1',), ('ix_datasets_id',)]`

So it looks like the composite index is successfully created when the database is first indexed.

ntellis:
I've tested this, and it's working perfectly.

My only thought is: does it make sense to check for this index when loading a database, ex in `PrecoveryDatabase.from_dir`? Because the performance really tanks without the index.

moeyensj:
That's a good idea. Basically, check for the index and see if it exists and if it doesn't then throw a warning?

ntellis (approved):
(Oops didn't actually submit the review)

Yes, that's what I was thinking. Simple warning. It's caught us in the past, when things have been slow, and hasn't been obvious.

I wouldn't call it a necessary change though, so approving either way.

moeyensj:
@ntellis I've added a warning when the index can't be found.

PR #43, opened 2022-10-31
Separate indexed datafiles into subdirectories of dataset and calendar months 

ntellis:
New file structure:

- data
   - NSC_DR2 (dataset_id)
      - 2013-07
         - Frames_00000001.data
         - …
      - …
   - …

Pre-commit hooks added a lot of reformatting, and should probably be reverted.

### DIFF
--- a/precovery/frame_db.py
+++ b/precovery/frame_db.py
@@ -4,22 +4,13 @@ import itertools
 import logging
 import os
 import struct
-from typing import (
-    Iterable,
-    Iterator,
-    Optional,
-    Set,
-    Tuple
-)
+from typing import Iterable, Iterator, Optional, Set, Tuple
 
 import numpy as np
 import sqlalchemy as sq
-from rich.progress import (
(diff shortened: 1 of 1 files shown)


precovery/frame_db.py
```diff
@@ -547,7 +546,12 @@ def load_hdf5(
             chunksize=chunksize,
         ):
             observations = [Observation.from_srcobs(o) for o in src_frame.observations]
-            data_uri, offset, length = self.store_observations(observations)
+            year_month_str = "-".join(
+                Time(src_frame.mjd, format="mjd", scale="utc").isot.split("-")[:2]
```
akoumjian, line 550: I think we could also do 
`Time(src_frame.mjd, format="mjd", scale="utc").isot.format_string("%YYYY-%MM")`

Not really important I just hate grabbing slices due to edge cases.

ntellis:
This PR should be made against a dev branch

PR #44, opened 2022-11-02
WIP: Docker, pure pip install

akoumjian:
Some possible changes:
* container builds for ease and portability of development, including compose to make mounting folders easier and uniform
* Attempt at simplifying dependencies by building oorb, ideally we can escape switching between conda and pip. We could also look at a PR for OORB that will build out python wheels with all the correct binaries and files (that's what wheels are for).
* I'd also like to use this as an opportunity to document the precovery input format and database format.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,2 +1,3 @@
 volume/
+data/
 .github/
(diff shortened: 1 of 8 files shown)


akoumjian:
Making this part of multiprocessing code.

PR #46, opened 2022-11-15
Ak/multiprocess indexing

akoumjian:


Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+RUN make
+RUN make ephem
+RUN cd /oorb/data && ./getBC430
+RUN cd /oorb/data && ./updateOBSCODE
+ENV LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/oorb/python/"
+ENV PYTHONPATH="$PYTHONPATH:/oorb/python/"
+ENV OORB_DATA="/oorb/data"
+ENV OORB_CONF="/oorb/main/oorb.conf"
+RUN wget https://www.minorplanetcenter.net/iau/MPCORB/MPCORB.DAT
+RUN /oorb/bin/oorb --task=mpcorb --mpcorb=MPCORB.DAT > /oorb/data/MPCORB.orb
```
moeyensj: Is this line to test if the oorb installation worked? The MPCORB is a fairly large file to download if we don't think we need it.

setup.py
```diff
@@ -1,8 +1,17 @@
 from setuptools import setup
 
 setup(
```
moeyensj: Does this break the semantic versioning we gain by using setuptools_scm?

requirements.txt
```diff
@@ -2,7 +2,7 @@ numpy
 numba
 sqlalchemy
 rich
-pytables
+tables
```
moeyensj: We may need to add a second version of the requirements file since we will want to support conda distribution of precovery as well. 

I think in conda its pytables and on pip its tables.

precovery/version.py
```diff
@@ -0,0 +1 @@
+__version__ = "0.1.0dev"
```
moeyensj: What are your thoughts on hardcoding the version? The semantic versioning available through `setuptools_scm` seems to be the Python standard for some of the most popular packages.

precovery/ingest/index_observations.py
```diff
@@ -0,0 +1,151 @@
...
+    dataset_id=None,
+    nside=DefaultConfig.nside,
+    data_file_max_size=DefaultConfig.data_file_max_size,
+    dataset_name=None,
+    reference_doi=None,
+    documentation_url=None,
+    sia_url=None,
+):
+
+    logger = multiprocessing.get_logger()
```
moeyensj, line 70: I didn't realize you could tap into multiprocessing's logger. This is cool!

precovery/ingest/index_observations.py
```diff
@@ -0,0 +1,151 @@
...
+
+    # Ensure database is initialized in main thread before starting
+    db = PrecoveryDatabase.create(
+        out_dir, nside=nside, data_file_max_size=data_file_max_size
+    )
+
+    time_start = datetime.datetime.now()
+
+    completed = 0
+    pool = multiprocessing.Pool(os.cpu_count() + 1)
```
moeyensj: For user-friendliness, I'd suggest adding `num_cpus` or similar to allow the user to say how many threads to use for indexing. This will be important for shared systems like epyc.

moeyensj (commented):
I'm looking forward to testing the indexing changes. This should really speed things up. The code for this component looks great. 

The only concern I have is the changes to the build system and versioning. I'm interested in hearing your thoughts. `pyproject.toml` and the build system that Python has transitioned to using [(PEP 518)](https://peps.python.org/pep-0518/) is something we should probably seek to maintain. I'm also personally a fan of semantic versioning based on git history over hardcoding versions, especially in the development context where it's easy to tell when an underlying file may have changed in an installed version of the code by looking at the version.

setup.py
akoumjian: I am happy to put this back. The issue I was running into was in using `gcloud build` to build the images which I was then using to run jobs. Gcloud build was failing because when it tars up the source to build the docker image remotely, it was failing due to the lack of the `.git` directory. I'm sure I can find another way around it.

precovery/ingest/index_observations.py
akoumjian: I think it's the only logger that works by default without a lot of headache.

akoumjian:
I'll put the scm versioning stuff back including the `pyproject.toml`.

ntellis:
Re: versioning. Would it be possible to track multiple version tags - a precovery version and an indexing version?

Thinking about a situation where it might be useful to write this explicitly to the database folder - i.e. keep the indexing version in the config.ini. Indexing ver could be incremented only with changes that affect indexing. Right now there is no way to tell in a db which version of precovery indexed it.

setup.py
moeyensj: Ah, that's interesting, that's a corner case I've never run into. I'd suspect some other Python packages will have worked around this before or something similar.

We could raise an exception and default to some version like astropy does: https://github.com/astropy/astropy/blob/main/astropy/version.py#L4

Dockerfile
akoumjian: I think something was failing without it (might have been a test of some kind?) I'll investigate.

akoumjian:
@ntellis I do think that databases should include what version built the database. I'm not sure I track the need for separate versions unless you were thinking that the library would only allow itself to run with a database built with an identical version. We could treat it like a client/server model.

setup.py
akoumjian: That might work. I'll definitely get this functioning again and figure it out one way or another.

precovery/frame_db.py
```diff
@@ -393,27 +385,41 @@ def all_frames(self) -> Iterator[HealpixFrame]:
...
-            exposure_id=frame.exposure_id,
-            filter=frame.filter,
-            mjd=frame.mjd,
-            healpixel=int(frame.healpixel),
-            data_uri=frame.data_uri,
-            data_offset=frame.data_offset,
-            data_length=frame.data_length,
-        )
-        self.dbconn.execute(insert)
+    def add_frames(self, frames: list[HealpixFrame]):
```
spenczar, line 408: Probably should be `frames: Iterable[HealpixFrame]`?

precovery/frame_db.py
```diff
@@ -547,7 +561,14 @@ def load_hdf5(
             chunksize=chunksize,
         ):
             observations = [Observation.from_srcobs(o) for o in src_frame.observations]
-            data_uri, offset, length = self.store_observations(observations)
+            year_month_str = "-".join(
+                Time(src_frame.mjd, format="mjd", scale="utc").isot.split("-")[:2]
+            )
```
spenczar, line 591: This feels like it deserves a more prominent billing, maybe as a method on the Frame class. Otherwise this is the only spot where it resides.

precovery/frame_db.py
```diff
@@ -561,32 +582,43 @@ def load_hdf5(
...
-        return "frames_{:08d}.data".format(self.n_data_files)
+            year_month_str = os.path.basename(os.path.dirname(abspath))
+            dataset_id = os.path.basename(os.path.dirname(os.path.dirname(abspath)))
+            data_uri = f"{dataset_id}/{year_month_str}/{name}"
+            if dataset_id not in self.n_data_files.keys():
+                self.n_data_files[dataset_id] = {}
+            if year_month_str not in self.n_data_files[dataset_id].keys():
+                self.n_data_files[dataset_id][year_month_str] = 0
+            self.data_files[data_uri] = open(abspath, "rb")
+            self.n_data_files[dataset_id][year_month_str] += 1
```
spenczar: This is complicated, and suggests that there might be a better data structure using something like a `Partition` class.

precovery/frame_db.py
```diff
@@ -561,32 +582,43 @@ def load_hdf5(
...
-            os.path.join(self.data_root, self._current_data_file_name())
+            os.path.join(
+                self.data_root, self._current_data_file_name(dataset_id, year_month_str)
+            )
         )
 
-    def _current_data_file_size(self):
-        return os.stat(self._current_data_file_name()).st_size
+    def _current_data_file_size(self, dataset_id: str, year_month_str: str):
+        return os.stat(self._current_data_file_name(dataset_id, year_month_str)).st_size
```
spenczar, line 645: It's pretty strange how the `dataset_id` and `year_month_str` are getting threaded through so many functions, and always as a pair. Is one ever provided separate from the other? I think this is evidence for a `PartitionIdentifier` as a dedicated concept.

precovery/frame_db.py
```diff
@@ -617,7 +651,14 @@ def store_observations(self, observations: Iterable[Observation]):
         which indicates the ID's length in bytes, followed by the UTF8-encoded
         bytes of the ID.
         """
-        f = self._current_data_file()
+        f = None
```
spenczar, line 681: Is this initialization necessary?

precovery/frame_db.py
```diff
@@ -617,7 +651,14 @@ def store_observations(self, observations: Iterable[Observation]):
...
+        try:
+            f = self._current_data_file(dataset_id, year_month_str)
+        except KeyError as ke:
```
spenczar: The implicit contract here - that `self._current_data_file` returns a KeyError if the datafile doesn't exist - feels a bit brittle. It's internal enough that it might not be worth fixing, but just worth giving a second thought.

precovery/ingest/index_observations.py
```diff
@@ -0,0 +1,157 @@
...
+    # --nside 32 --dataset_name "NOIRLab Source Catalog (DR2)" --reference_doi https://doi.org/10.3847/1538-3881/abd6e1
+    # --documentation_url https://datalab.noirlab.edu/nscdr2/index.php --sia_url https://datalab.noirlab.edu/sia/nsc_dr2
+    logger.info(f"Found {len(files)} observation files in {data_dir}:")
+    for f in files:
+        logger.info(f"\t{os.path.basename(f)}")
+
+    # Ensure database is initialized in main thread before starting
+    db = PrecoveryDatabase.create(
+        out_dir, nside=nside, data_file_max_size=data_file_max_size
+    )
```
spenczar, line 81: perhaps db.,close() right after initializing? On my first read, I was expecting it might get used by the pool; closing it immediately would make it clear that doesn't happen.

precovery/ingest/index_observations.py
```diff
@@ -0,0 +1,157 @@
...
+
+    time_start = datetime.datetime.now()
+
+    completed = 0
+    cpu_count = args.cpu_count or multiprocessing.cpu_count() + 1
+
+    pool = multiprocessing.Pool(cpu_count)
+
+    for result in pool.starmap(
```
spenczar: This is wonderfully simple. Very nice.

spenczar (commented):
I think that this code would benefit from a mild change to the data model. Adding a new `Partition` class could help contain the partitioning logic and simplify the FrameDB's methods.

The Partition class could be responsible for:
 - Constructing and unpacking filepath strings
 - Checking if a partition already exists on disk
 - Creating new partitions
 - Keeping track of the `n_data_files` within a partition

In addition, a simple `PartitionKey` class, which is really just a tuple of the partitioning keys (dataset ID and year-month) would help.
 
FrameDB could keep a collection of partitions, which is a dict of PartitionKeys to Partitions. SourceFrames could have a method to produce a PartitionKey.

This general approach would pull things out of strings and dicts and into top-level objects, which I think will make changes to the partitioning scheme easier. I think it might make it easier to read, too.

I suspect that this partitioning (by month) is quite skewed, and we may want to try different partitionings. Maybe it doesn't matter since the partitioning is only relevant (I think?) during indexing, and doesn't really affect read performance, but up to you.

Of course, since I'm not actively in the project, take all of these as suggestions; I think the current thing clearly would work too.

precovery/frame_db.py
spenczar: Ah, I reveal that my `typing` knowledge is from before Python 3.9: https://docs.python.org/3/whatsnew/3.9.html#type-hinting-generics-in-standard-collections

ntellis:
@spenczar Thanks for the excellent review.

I like the scheme you propose for pulling out the partitioning information - the string logic is messy and the `n_data_files` dict felt like an unclear abstraction.

Re: monthly partitioning. The main motivation for this was easily supporting parallelism - give threads non-overlapping sets of source files to index, and they won't be writing to the same binary files. But it does have its limitations (namely, each month being processed by at most one thread at a time).

precovery/frame_db.py
ntellis: Agreed completely. Maybe it would make more sense for `_current_data_file` to have a create=True flag, create a new datafile if necessary, and avoid the KeyError nonsense altogether.

requirements.txt
akoumjian: I will update the github actions to build both for pypi and conda, as well as documentation.

precovery/frame_db.py
akoumjian: Agreed. Calendar-month from mjd is also implemented in `adam-etl` already as well. This would benefit from being somewhere in our new theoretical `adam.core`.

akoumjian:
@ntellis Did you want me to handle the Partition class suggestions? I'm happy to.

ntellis:
@akoumjian If you're already in there and feel that we should pull out that logic, sure. I'd be happy to do so too.

Dockerfile
akoumjian: I removed it and I believe it's working, still.

requirements.txt
akoumjian: I'll be pushing for the pypi package to simply use `setup.cfg` and for the conda install to use what's specified in `meta.yml` if I'm understanding how it works, correctly. I don't believe we need the requirements files at all and this way we have distinct sets of requirements.

akoumjian:
@ntellis I haven't gotten to the partitioning yet, still working on the build / pipeline stuff which I think is 90% there. I'd like to discuss what the bare minimum is necessary to get this merged in just for the sake of shorter lived branches and forking off this to adding the new unit tests.

akoumjian:
Going to merge this sans the partition class just so we have a working frame of reference. Our new test is passing, afterall.

PR #52, opened 2022-12-20
Suggest moving to csv

akoumjian:
Moving to CSV as our base serialization format will have several benefits over the current h5.

Pros
* Much less disk usage (for the size of our files, we are not gaining anything from h5 compression)
* Much simpler file append mechanics
* File generation is deterministic, which means we can use hashing functions to quickly check for changes. This will let us efficiently do caching comparisons between all the different steps. Weirdly, h5 files are not bit deterministic on disk
* Easy visual inspection of serialized values on disk, useful for checking float precision and other errors

Cons
* We will lose the ability to use `store.select` for on the fly SQL - like analysis. This really only affects individual files larger than our machine RAM limits, otherwise we just read it into a dataframe in memory. We don't have too many of those since we break up the files by year-month (largest NSC is 2.2GB).
* we have some existing `.h5` files that we like to reference / use. I've preserved ingesting by h5 in this PR, so we don't really lose that.


tests/test_precovery_index.py
```diff
@@ -17,7 +17,7 @@ def test_precovery():
         out_dir="database",
         dataset_id="test_dataset",
         dataset_name="Test Dataset",
-        data_dir="tests/data",
+        data_dir="tests/data/index",
```
moeyensj: If you rebase with main now there will be a merge error on these paths. I've been using absolute paths to make sure the test files always stay in the test directory. Not sure what the best practice is but if you adopt that then I think we can update the .gitignore to ignore some of the generated files in the test directory.

tests/test_precovery_index.py
```diff
@@ -44,8 +44,8 @@ def test_precovery():
         )
         orbits_keplerian.append(orbit)
 
-    # Load observations from h5 file
-    observations_df = pd.read_hdf(TEST_OBSERVATION_FILE)
+    # Load observations from csv file
+    observations_df = pd.read_csv(TEST_OBSERVATION_FILE)
```
moeyensj: This might be a bit excessive but do we want to test the behavior between h5-based indexing and csv-based indexing since we are maintaining both? If so, we could add a unit test that indexes the observations from h5 and csv and then does a comparison on the output directory. Likely overkill so please don't feel like this needs to happen for this to be merged.

moeyensj (approved):
This looks great! 

I believe the main reason we swapped to HDF5 was that for large files it takes pandas a lot longer to read in a CSV than it does an HDF5 file. We inherited some catalogs of data in this format from simulations work with LSST. However, at the scale we are now talking about with the multiprocessed and time-based indexing implemented, I think the benefits of having CSVs definitely far outweight the benefits of HDF5.

tests/test_precovery_index.py
akoumjian: While this PR maintains both, I would prefer to keep it simple and only go with the one file format. If we did maintain both, then yes I would definitely add a test.

tests/test_precovery_index.py
akoumjian: Yeah, I have no problem making those changes. I'm trying to recall the reason I switched this up and I think it had something to do with scanning for files. Didn't want the sample orbits file getting pulled in along with the to-be-indexed observations file. Previously this didn't conflict because they had separate extensions and so the folder scanning didn't pick up the false positive when they lived in the same spot.

akoumjian:
@moeyensj I've rebased off your latest schema changes. It's failing on the test, I believe due to reasons you already solved. Am I missing something in the build (reference to version of pyoorb?) that would fix the automated tests?

tests/make_observations.py
```diff
@@ -227,4 +227,4 @@ def make_observations(
     args = parser.parse_args()
     orbits_df = pd.read_csv(args.in_file)
     observations = make_observations(orbits_df, orbit_type=args.orbit_type)
-    observations.to_hdf(args.out_file, key="data", mode="w", format="table")
+    observations.to_csv(args.out_file, index=False)
```
moeyensj: I suspect this is the reason the test is failing. I've run into issues in the past with pandas serialization of floats and our tests are fairly aggressive at the moment. We may need to define the `float_format` argument here or reduce the test tolerance. If you set float_format to "%.16f", what happens?

akoumjian: I tried both `float_format` in generating the observations as well as `float_precision="high"` when reading them in for comparison. Unfortunately all combinations of both still failed. I think we've got a precision modification happening during indexing itself that I'll check tomorrow.

akoumjian: Okay, looks like `float_precision="round_trip"` does it when reading the test observation csv directly fixed the issue. Pandas is heavily optimized for read speed rather than precision for csv floats apparently.

PR #55, opened 2023-01-09
Make precovery install editable in Dockerfile

moeyensj:
Make the precovery installation editable to avoid having to rebuild the image when changes are made to the code.


akoumjian (approved):
Looks good, and remember if you use it with docker compose you automatically get your folder mounted to the directory that is symlinked with `-e`.

PR #56, opened 2023-01-10
Add per-observation MJDs [ADAM-70]

moeyensj:
To address the LSST use case, this PR adds per-observation MJDs to the binary files. In instances where the observation is not measured at the midpoint of the exposure then a 2-body propagation is carried out to update the predicted location of the orbit.

Note:
- This PR removes the approximately_propagate and the linear propagation functions in favor of making 2-body the standard for approximate propagations 
- The tolerance for the synthetic indexing test has been increased to account for errors introduced by approximate propagation (tests are now within a milliarcsecond: this is still 10-100x larger than the errors we expect in astrometry from LSST)
- "mjd_start", "mjd_mid" are renamed to "exposure_mjd_start", "exposure_mjd_mid" internally
- For inputs to the indexing code and for the results returned by precovery "_utc" is appended the above to explicitly state the time scale (so "exposure_mjd_start_utc", "exposure_mjd_mid_utc")


precovery/frame_db.py
```diff
@@ -247,8 +250,8 @@ def get_frames(
         ).where(
             self.frames.c.obscode == obscode,
             self.frames.c.healpixel == int(healpixel),
-            self.frames.c.mjd_mid >= mjd - 1e-7,
-            self.frames.c.mjd_mid <= mjd + 1e-7,
+            self.frames.c.exposure_mjd_mid >= mjd - 1e-7,
+            self.frames.c.exposure_mjd_mid <= mjd + 1e-7,
```
spenczar, line 254: This doesn't seem correct anymore for multi-timestamped frames. It seems to me we want any where mjd is within the exposure duration. That is, something like:
```python
self.frames.c.exposure_mjd_start <= mjd - 1e-7,
self.frames.c.exposure_mjd_start + self.frames.c.exposure_duration / 86400 >= mjd + 1e-7
```
but this might not be quite right for ones with a single timestamp for all observations? I'll let you do the hard thinking on that .

precovery/frame_db.py
```diff
@@ -479,15 +486,17 @@ def initialize_tables(self):
...
             sq.Column("exposure_duration", sq.Float),
             sq.Column("healpixel", sq.Integer),
             sq.Column("data_uri", sq.String),
             sq.Column("data_offset", sq.Integer),
             sq.Column("data_length", sq.Integer),
-            # Create index on mjd, healpixel, obscode
-            sq.Index("fast_query", "mjd_mid", "healpixel", "obscode", unique=True),
+            # Create index on midpoint mjd, healpixel, obscode
+            sq.Index(
+                "fast_query", "exposure_mjd_mid", "healpixel", "obscode", unique=True
```
spenczar, line 498: Note that if my previous suggestion is implemented, this needs adjustment.

precovery/precovery_db.py
```diff
@@ -377,43 +386,86 @@ def _check_frames(
...
                 n = len(obs)
+
+                # Read observation mjds, ra, decs into arrays
+                obs_mjds = np.array([o.mjd for o in obs])
                 obs_ras = np.array([o.ra for o in obs])
                 obs_decs = np.array([o.dec for o in obs])
+
+                # If any observation MJD does not match the mjd of the ephemeris
+                # to within 1e-8 days or ~1 ms then adjust the
+                # predicted ephemeris accordingly
```
spenczar: This doesn't feel like quite the right condition. Isn't it more like "if the frame contains multiple different observation timestamps?" Or perhaps "if it has a duration > 0"?

I realize they'll often (maybe even always) be the same, but using this more explicit condition would be clearer, I think.

precovery/precovery_db.py
```diff
@@ -424,17 +476,17 @@ def _check_frames(
                 logger.info("checked %d observations in frame", n)
                 if (len(obs) == 0) & (include_frame_candidates):
```
spenczar: While you're here, may as well fix this - it should be `and` not `&` in there. The bitwise-and is only working because of dumb luck, right now.

precovery/sourcecatalog.py
```diff
@@ -142,75 +143,80 @@ def iterate_observations(
...
                 "dec",
                 "ra_sigma",
                 "dec_sigma",
                 "mag",
                 "mag_sigma",
                 "filter",
-                "mjd_start_utc",
-                "mjd_mid_utc",
+                "exposure_mjd_start_utc",
+                "exposure_mjd_mid_utc",
```
spenczar: I don't really love that there is a `_utc` suffix here, but not elsewhere in the code. What's the reason for that symmetry?

tests/make_observations.py
```diff
@@ -13,6 +13,8 @@
 TEST_OBSERVATION_FILE = os.path.join(
     os.path.dirname(__file__), "data", "observations.h5"
 )
+RANDOM_SEED = 2023
+np.random.seed(RANDOM_SEED)
```
spenczar: To be pedantic for a second:

This is a little dodgy because it modifies `np.random` for _all_ tests within the same Python interpreter run. So other tests will also get your seed - or you might get theirs. For almost all practical purposes, that should be fine, but if we had any "golden" tests (that is, tests that a particular output was unchanged across versions, like "gold plates") which involved randomness, this could break them in a very hard-to-expect way. 

An alternative strategy which wouldn't have this problem would be to make a new Numpy `Generator` within your test:

```python
rng = np.random.default_rng(seed=2023)
offset_from_start = rng.uniform(dts, dts + exposure_duration / 86400)
```

This would insulate things. Almost definitely not necessary, but hey, why review code if you don't get to pass in pedantic commentary like this?

tests/make_observations.py
```diff
@@ -137,10 +139,12 @@ def make_observations(
...
-        observation_times = (
-            initial_epoch.utc.mjd + dts + exposure_duration / 86400 / 2.0
-        )
+
+        # Calculate a random offset from the start of the exposure
+        # to give each observation a unique obervation time
+        offset_from_start = np.random.uniform(dts, dts + exposure_duration / 86400)
+        observation_times = initial_epoch.utc.mjd + offset_from_start
+        exposure_ids = [f"{obs_i}_{k:06d}" for k, obs_i in enumerate(observatory_codes)]
         exposure_ids = [f"{obs_i}_{k:06d}" for k, obs_i in enumerate(observatory_codes)]
```
spenczar: This line looks duplicated.

tests/make_observations.py
```diff
@@ -137,10 +139,12 @@ def make_observations(
...
-        # Observation times are defined at the center of the exposure (for now)
-        observation_times = (
-            initial_epoch.utc.mjd + dts + exposure_duration / 86400 / 2.0
-        )
+
+        # Calculate a random offset from the start of the exposure
+        # to give each observation a unique obervation time
+        offset_from_start = np.random.uniform(dts, dts + exposure_duration / 86400)
+        observation_times = initial_epoch.utc.mjd + offset_from_start
+        exposure_ids = [f"{obs_i}_{k:06d}" for k, obs_i in enumerate(observatory_codes)]
```
spenczar: Why is this inside the loop over `orbits`? It seems like it should be done once for all orbits during the initial setup.

spenczar (changes requested):

precovery/frame_db.py
moeyensj: Correct me if I'm wrong here. What we want this function to do is to get all the frames for a given midpoint MJD and obscode (not the observations themselves). The query here is against the frames table so I believe this should work to pull all unique frames for a given exposure time since all the frames built from one exposure will have the same exposure_mid_mjd. We don't need exposure duration since exposure_mid_mjd and obscode uniquely define the collection of frames and by extension the observations contained within the frames.

precovery/sourcecatalog.py
moeyensj: I agree. I don't like it either. I originally added "_utc" to the user-facing inputs and outputs to be explicit about the time scale in which we expect the times to be defined because time scales are always something that bite us somewhere. I'll remove it now because I believe the README is fairly clear on the expected time scale for the inputs. I'll add a description of our outputs to make it clear we are returning times in UTC as well.

tests/make_observations.py
moeyensj: No particularly good reason other than it seemed better to add different offsets for different orbits. In other words, I wanted each orbit to have unique dts to complement the fact that they already are not constrained to have the same initial epochs (by design). In the event, however, that they do have the same initial epoch then I wanted one additional layer to separate their observation times.

precovery/frame_db.py
spenczar: Hm, yeah, this is a bit complicated. Let me try and chatter out loud about what's going on here:

We've propagated the orbit(s) to our "propagation targets" (https://github.com/B612-Asteroid-Institute/precovery/blob/5f944ff6ec0000ae8a930a1a8577c88e02d70d96/precovery/frame_db.py#L191) which are the unique *exposure* MJDs in the frames table within a window.

Then we're filtering to frames that match the healpixel *and* the exposure MJD, and here we're checking each individual source to see if it matches up with our ephemeris. We do that by propagating each source back to the exposure midpoint as if it were on our candidate orbit.

So I think you're right that when we're selecting frames that match, it's just about the midpoint.

---

I am finding four things confusing when trying to talk this out:
1. It takes a lot of mental effort for me to keep track of whether we're discussing the MJD of an exposure, or of an observation, or of an ephemeris. Naming variables better will help here, probably. A name like `mjd` is not really sufficiently descriptive.
2. Many of the functions in `frame_db` have gotten too big. `_check_frames` is over 100 lines long! These need to be broken up into pieces.
3. MJDs are floats, which is really nasty to work with in a database, so we have to do all these approximate checks. Maybe we could convert to storing UTC nanoseconds, and only work in MJD when we're doing orbital math. This is only valid if the MJD <-> UTC conversion is trivial math.
4. It's tricky that we're storing the start _and_ midpoint _and_ duration of exposures. Just two of those would describe the third value. And here, we are only using the midpoint. Can you remind me why we need duration/start?

Not all of those should be fixed here, or perhaps any of them. Just recording them.

precovery/frame_db.py
spenczar: As @moeyensj pointed out above, my suggestion wasn't correct.

tests/make_observations.py
spenczar: That's not how real data would work, though, right? Wouldn't the observations always have consistent timestamps, even for orbits with the same or different epochs?

precovery/frame_db.py
akoumjian: Here's my personal take on each of these.

1. This is part of my motivation of exploring backend db alternatives where we aren't limited to the hard coded indices of the frames. Not sure what else will solve this other than lengthy comments in the code making sure each step has a narrative context. I personally don't mind long variable names.
2. I think maybe @ntellis has some work that does some of this, but I could be wrong. Agree on the refactoring.
3. I agree with this. Using types with unambiguous serialization avoids so many problems. Reminds me a lot of [unicode sandwich](https://nedbatchelder.com/text/unipain/unipain.html#1). The downside with not using floats is that your precision is fixed. If we have a known precision that is the limit we care about, it's not a problem.
4. This is what I was trying to say during the live review. I kind of want an opinionated version on which ones to use. Each value we store and maintain is room for error to be introduced.

tests/make_observations.py
moeyensj: Pedantisms appreciated and incorporated. A much cleaner change and I've learned about a new `numpy` feature.

precovery/sourcecatalog.py
moeyensj: @spenczar @akoumjian Would you want me to also rename "mjd_utc" to "mjd" for consistency? (These are the observation times).

tests/make_observations.py
moeyensj: It's unlikely two moving objects observed in the same exposure will have the same time unless they were exactly positioned along the axis perpendicular to the motion of the shutter so instead this synthetic data doesn't model the shutter and just makes it random for each orbit within the confines of an exposure. 

I think what it comes down to is that the synthetic data isn't modeling the footprint of images on the sky. The sample orbits are all over the place and so two orbits observed on the sky may be at polar opposite locations but still be within the same "exposure". We are really just testing that given an orbit and some dummy observation times do we extract all of the observations with the input orbit? The times, the cadences, observatory codes, exposure durations, etc.. are all potential variables on how this test may perform but they may not represent reality for any given set of observations from a real telescope.

precovery/sourcecatalog.py
spenczar: Yes, please. I think naming it `observation_time` or something similar that indicates its describing the observation may make sense, but that's okay to do later.

precovery/frame_db.py
moeyensj: Regarding number 1. We could prepend "obs_" for all observation quantities and "exp_" for all exposure quantities. I tradtionally prepend "pred_" for anything ephemeris related.

precovery/sourcecatalog.py
moeyensj: Okay, I'll swap it to "mjd" for now. We can do code-wide variable name changes in a later PR.

tests/make_observations.py
spenczar: Ah! You aren't making a simulated *exposure*, you're just generating the ephemerides for some orbits, and then casting 'em back into simulated "observations." Got it. Of course, that's what the function's name and docstring indicates. My mistake.

Yes, your arguments all make sense for using random offsets for each orbit.

precovery/frame_db.py
spenczar: obs and exp are alright with me, although I agree with Alec that long variable names are totally fine. "exp" is sometimes a prefix for "experimental" or "exponential" so it takes extra thinking, for example, compared to "`exposure_`". 

What is 'pred'? Is it "predicted"? If so, `predicted_` seems like a better prefix, or even better, `ephemeris_`.

precovery/precovery_db.py
moeyensj: Implemented as you suggested just compared against 0.

moeyensj:
@spenczar If I did the rebase correctly, I think I addressed all your comments. Ready for another pass!

precovery/frame_db.py
```diff
@@ -309,28 +314,28 @@ def frames_for_bundle(
...
-            sqlfunc.max(self.frames.c.mjd_mid, type=sq.Float),
+            sqlfunc.min(self.frames.c.exposure_mjd_mid, type=sq.Float),
+            sqlfunc.max(self.frames.c.exposure_mjd_mid, type=sq.Float),
         )
         first, last = self.dbconn.execute(select_stmt).fetchone()
         return first, last
 
     def frame_bundles(
-        self, window_size_days: int, mjd_start: float, mjd_end: float
+        self, window_size_days: int, exposure_mjd_start: float, mjd_end: float
```
spenczar: I don't think renaming this one to `exposure_mjd_start` makes sense.

spenczar (approved):
One little thing, but seems good to me.

precovery/frame_db.py
moeyensj: Ah yes, good catch. Let me fix it and then I'll merge.

PR #57, opened 2023-01-20
Install pyoorb from binary wheel

spenczar:


Dockerfile
```diff
@@ -1,28 +1,30 @@
...
-ENV OORB_CONF="/oorb/main/oorb.conf"
-RUN make install
+# Install oorb data
+RUN curl -L -o /tmp/oorb_data.zip \
+    https://github.com/B612-Asteroid-Institute/oorb/releases/download/$OORB_TAG/oorb_data.zip
+RUN unzip -d /opt/oorb_data /tmp/oorb_data.zip
+ENV OORB_DATA=/opt/oorb_data
+
+# Install pyoorb
+RUN export WHEEL_NAME="${OORB_VERSION}-cp310-cp310-manylinux_2_17_$(uname -m).manylinux2014_$(uname -m).whl" && \
```
akoumjian, line 22: Is there a reason we need these dynamically assigned in the Dockerfile instead of added to the `setup.cfg` for precovery? I'm guessing this is to allow precovery to be installed in multiple environments and architectures combined with the fact that a `.whl` on Github can't be selected dynamically like it can from a package server.

akoumjian (approved):
Looks good. What should the evolution from here be? Does it require getting both the `.whl` building PR and potentially our fork of `pyoorb` getting merged in? Or should we leave as is since we may be evaluating OD library usage long term.

akoumjian:
Obviously we also need to figure out why the tarball is not valid in the CI.

Dockerfile
spenczar: Yeah, there are two things set here. 

One is the version of `pyoorb`. This could be done in setup.cfg, but since our required version is not published to PyPI, a simple 'pip install' would not work for anyone; they'd need to manually download the wheel like this.

The second is the platform name. This is a feature of the machine, and so it couldn't be set by `setup.cfg`. We could fix it to always be x86_64 by making the first line of this `FROM --platform=linux/amd64 ubuntu:jammy`, which is a strange-but-valid syntax. But then we'd be forcing emulation when running on M1 Macs, and emulation is very slow. I would prefer to build wheels for all target architectures and install the right wheel for the host machine.

Even if we set the pyoorb version in `setup.cfg` and relied on the `--platform` tag in the Dockerfile, we still need to figure out this big wheel filename so we can download the right thing from Github, like you say. The only way around this is to index our wheels on a Python Package Index. We could do that on public PyPI, but with pyoorb we're actually building a fork. I don't want to claim the public `pyoorb` name ahead of github.com/oorb/oorb. 

So that leaves a private PyPI, which probably proxies public packages to https://pypi.org. That would simplify this installation, and it's not nearly as hard as it may sound, but it's more work than I wanted to do right here.

spenczar:
> What should the evolution from here be? Does it require getting both the .whl building PR and potentially our fork of pyoorb getting merged in? Or should we leave as is since we may be evaluating OD library usage long term.

Yes, I think we have two possible futures:
 - run a private PyPI with our packages, including modified forks, or
 - get our PR merged in and get oorb to publish to PyPI

Dockerfile
akoumjian: > So that leaves a private PyPI, which probably proxies public packages to https://pypi.org/. That would simplify this installation, and it's not nearly as hard as it may sound, but it's more work than I wanted to do right here.

Okay, that lines up with my understanding. We will maintain the "install pyoorb separately" workflow for now and push for merge into the project and publishing the updates. 

The only other work involved then is a well developed fix for the pyoorb failure mode to PR into main.

spenczar:
> Obviously we also need to figure out why the tarball is not valid in the CI.

This is resolved, I expanded the shell variable the wrong way in the curl string. Added `-f` to the flags so it errors on a 404 (which this was) instead of downloading a 404 HTML page and trying to unzip it :]

PR #58, opened 2023-01-24
Add multi-dataset synthetic test data and unit test [ADAM-47]

moeyensj:
This PR extends the synthetic test data generation to generate multiple different datasets of observation spanning about 2 calendar months with overlapping and non-overlapping regions of coverage. 

From the `tests/README.md`:
```
#### `make_observations.py`:
Produces 4 synthetic observations CSV files containing observations for each object in the `sample_orbits.csv` file. The observations have the following properties:
- 4 observations each day (a "quad" tracklet) for 15 days for each observatory.
- Each observatory ('500', 'I11', 'I41', 'F51') is split into its own unique dataset.
- Each observatory has an observation window that starts at 0, 10, 20, and, 30 days after the initial epoch at which the orbits are defined for '500', 'I11', 'I41', 'F51', respectively. This, by design, creates periods of overlap and non-overlap for each dataset and spreads observations over a 2 calendar month period.
- The first observation on any night is always an integer number of days from the reference epoch at which the orbit is defined plus a ~few hours time shift to account for the location of the observatory relative to GMT. Observations 2, 3, and 4 in the same night as the first, are set to be 30, 60, and 90 minutes from the time of the first observation, respectively.
- The exposure times increase for each observation in the quad tracklet: 30, 60, 90, and 120 seconds.
- **No** astrometric errors are added. 
- The observation times are reported at a random time between the start of the exposure and the end of the exposure.  

Usage: ```python make_observations.py```  
This saves observations to `data/index/dataset_500/dataset_500_observations.csv`, `data/index/dataset_I11/dataset_I11_observations.csv`, `data/index/dataset_I41/dataset_I41_observations.csv`, `data/index/dataset_F51/dataset_F51_observations.csv`.
```


tests/test_precovery_index.py
```diff
@@ -58,27 +50,51 @@ def test_precovery(test_db_dir):
...
+
+        index(
+            out_dir=test_db_dir,
+            dataset_id=dataset_id,
+            dataset_name=dataset_id,
+            data_dir=os.path.join(
+                os.path.dirname(__file__), f"data/index/{dataset_id}/"
+            ),
+            nside=16,
+            cpu_count=1,
```
moeyensj: `cpu_count=1` is a residual from when I disabled multi-processing during testing. I'll remove this before merging and set it back to default.

tests/README.md
```diff
@@ -13,15 +13,16 @@ Usage: ```python get_orbits.py```
 This saves orbits to `data/sample_orbits.csv`.
 
 #### `make_observations.py`:
-Produces a synthetic observations CSV file containing observations for each object in the `sample_orbits.csv` file. The observations have the following properties:
-- 4 observations each day (a "quad" tracklet) for 15 days from the epoch at which the orbit is defined (60 observations total).
-- Each set of 15 observations (in order of increasing time) is made by a different observatory: "500", "I11", "I41", "F51".
-- The first observation on any night is always an integer number of days from the reference epoch at which the orbit is defined. Observations 2, 3, and 4 in the same night, are set to be 30, 60, and 90 minutes from the time of the first observation, respectively.
+Produces 4 synthetic observations CSV files containing observations for each object in the `sample_orbits.csv` file. The observations have the following properties:
+- 4 observations each day (a "quad" tracklet) for 15 days for each observatory.
+- Each observatory ('500', 'I11', 'I41', 'F51') is split into its own unique dataset.
```
akoumjian: Unlikely to cause a but, but might make more sense to collapse one of these to have multiple observatories in a single dataset so it resembles datasets like NSC?

akoumjian (approved):

tests/README.md
moeyensj: I think that's a great idea to be on the safe side. I'll rebase and then combine the F51 and I41 observations into a single dataset.

moeyensj:
I've combined I41 and F51 into a single dataset and updated the tests README accordingly. The `cpu_count=1` testing residual has also been reverted.

PR #59, opened 2023-01-25
Update pre-commit

spenczar:
Updated pre-commit to use a modern version of black, since the old pinned version has a version compatibility issue with the latest version of click.

Removed pip-tools from pre-commit since it appears they are no longer used.

Ran pre-commit run --all-files, which made a lot of whitespace changes.


moeyensj:
One minor issue for Markdown files is that two whitespaces at the end of a line are interpreted as wanting a line break so if we clear trailing whitespaces there we lose formatting capabilities. I've set my vscode to ignore removing whitespaces in *.md files on save.

spenczar:
> One minor issue for Markdown files is that two whitespaces at the end of a line are interpreted as wanting a line break so if we clear trailing whitespaces there we lose formatting capabilities. I've set my vscode to ignore removing whitespaces in *.md files on save.

Whoa, that's news to me. Very strange.

Do we actually lose formatting capabilities, or we'd need to format line breaks with actual line breaks?

spenczar:
https://github.com/pre-commit/pre-commit-hooks/pull/58 appears to fix this. I'll make a change to the config (possibly tomorrow) to specifically ignore markdown soft breaks.

moeyensj:
> > One minor issue for Markdown files is that two whitespaces at the end of a line are interpreted as wanting a line break so if we clear trailing whitespaces there we lose formatting capabilities. I've set my vscode to ignore removing whitespaces in *.md files on save.
> 
> Whoa, that's news to me. Very strange.
> 
> Do we actually lose formatting capabilities, or we'd need to format line breaks with actual line breaks?

We could swap to using `<br>` with a caveat. I've had this debate in my mind before and ended up consulting: https://www.markdownguide.org/basic-syntax/#line-break-best-practices Unfortunately, not all Markdown applications support HTML tags though.

moeyensj:
> [pre-commit/pre-commit-hooks#58](https://github.com/pre-commit/pre-commit-hooks/pull/58) appears to fix this. I'll make a change to the config (possibly tomorrow) to specifically ignore markdown soft breaks.

Nice find! Yeah, this would do it.

spenczar:
@moeyensj Okay, fixed it, and restored the old softbreaks.

moeyensj:
Regarding the actual purpose of this PR: I'm more than happy to adopt pre-commit. The main reason we dropped it is as you suspected: you were the only one using it. If we agree on the formatting org-wide then I'm in favor of adding pre-commit as a development dependency since it will make our contributions easier to parse. We will want to update [our developer guide](https://github.com/B612-Asteroid-Institute/developer-wiki/wiki/Developer-Guide) with the relevant details if we decide to merge. 

I'm leaving that decision to our head of software engineering!

akoumjian (changes requested):
I'm all for this, however I've never actually used pre-commit hooks before. Can you include both in the README and in the developer guide Joachim mentioned some instructions on how to set it up properly and any recommended environment setups?

spenczar:
@akoumjian I've added README.md instructions. I've also written a little wiki page here: https://github.com/B612-Asteroid-Institute/developer-wiki/wiki/Using-pre-commit

I can merge that into the Developer Guide on the wiki if you think it looks good.

.github/workflows/lint-test.yml
```diff
@@ -17,8 +17,12 @@ jobs:
...
         continue-on-error: true
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+        with:
+          install: true
       - name: Build Precovery in Docker
-        run: docker build -t precovery:$IMAGE_TAG .
+        run: docker build --load -t precovery:$IMAGE_TAG .
       - name: isort check
         run:  docker run -i precovery:$IMAGE_TAG isort --check-only .
```
akoumjian: These are going to fail because we removed the packages from `setup.cfg`. Instead, it now relies on `pre-commit` running (I don't think dependencies get installed until you perform a run?). So we can reduce all these steps to a single `pre-commit run --all-files` or we can do something else.

akoumjian (approved):
Just need to update the actions to also use precommit per the other comment.

.github/workflows/lint-test.yml
spenczar: Yep, good call, those should be removed. The pre-commit dependencies get installed in a separate virtualenv anyway so they aren't accessible in this way.

spenczar:
Linting failures are fixed in https://github.com/B612-Asteroid-Institute/precovery/pull/61.

PR #60, opened 2023-01-25
Add functions for adding data to FrameDB

spenczar:
Added functions for adding datasets and SourceFrames directly to a FrameDB.

Rearranged how data partitition timestamps are computed for clarity.

Added tests.

### DIFF
--- a/precovery/frame_db.py
+++ b/precovery/frame_db.py
@@ -642,15 +642,9 @@ class FrameDB:
         dataset_id: str,
         skip: int = 0,
         limit: Optional[int] = None,
-        name: Optional[str] = None,
-        reference_doi: Optional[str] = None,
-        documentation_url: Optional[str] = None,
-        sia_url: Optional[str] = None,
     ):
         """
         Load data from a CSV observation file.
-
-        Creates a new dataset if the given dataset_id is not already present.
 
         Parameters
         ----------
@@ -663,16 +657,7 @@ class FrameDB:
             Number of frames to skip in the file.
         limit : int, optional
             Maximum number of frames to load from the file. None means no limit.
-        name : str, optional
-            User-friendly name of the dataset.
-        reference_doi : str, optional
-            DOI of the reference paper for the dataset.
-        documentation_url : str, optional
-            URL of any documentation describing the dataset.
-        sia_url : str, optional
-            Simple Image Access URL for accessing images for this particular dataset.
         """
-        self.add_dataset(dataset_id, name, reference_doi, documentation_url, sia_url)
         frames = sourcecatalog.iterate_frames(
             csv_file,
             limit,
--- a/precovery/frame_db.py
+++ b/precovery/frame_db.py
@@ -541,6 +541,101 @@ class FrameDB:
         for f in self.data_files.values():
             f.close()
 
+    def add_dataset(
+        self,
+        dataset_id: str,
+        name: Optional[str] = None,
+        reference_doi: Optional[str] = None,
+        documentation_url: Optional[str] = None,
+        sia_url: Optional[str] = None,
+    ):
+        """Add a new (empty) dataset to the database.
+
+        If a dataset with the provided dataset_id already exists,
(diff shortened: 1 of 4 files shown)


precovery/frame_db.py
```diff
@@ -626,9 +620,66 @@ def load_csv(
...
             )
-            frames_to_add.append(frame)
 
-        self.idx.add_frames(frames_to_add)
+            healpix_frames.append(frame)
+
+        self.idx.add_frames(healpix_frames)
+
+    @staticmethod
+    def _compute_year_month_str(frame: sourcecatalog.SourceFrame) -> str:
```
akoumjian, line 629: This is unimportant, but maybe this should live as a staticmethod on `sourcecatalog.SourceFrame`? Since you're not passing in more general types here (datetime, date string, etc.), the internal knowlege about `SourceFrame` feels weirdly coupled. I guess that's true for `add_frames` as well and maybe the whole `sourcecatalog` module is really just a sub of `frame_db` since I don't think we really use it anywhere else?

precovery/frame_db.py
```diff
@@ -626,9 +620,66 @@ def load_csv(
...
+        return time.strftime("%Y-%m")
+
+    def load_csv(
+        self,
+        csv_file: str,
+        dataset_id: str,
+        skip: int = 0,
+        limit: Optional[int] = None,
+        name: Optional[str] = None,
+        reference_doi: Optional[str] = None,
```
akoumjian: You know, I think even with this utility it's totally reasonable to split out the `add_dataset` function. If we want to index files from the cli, our cli function can take the command line arguments and make separate calls to `add_dataset` and `load_csv`. If we're calling these methods programmatically (say, inside an ETL process), then it's just as good to be able to make separate calls. That way, the simplest case remains simple (loading a single file from cli) but we don't get this weird thing where loading 20 csv's for a large dataset tries to upsert the dataset info 20 times.

tests/test_frame_db.py
```diff
@@ -38,3 +57,191 @@ def test_fast_query_warning(test_db):
...
+        exposure_mjd_start=t_start.mjd,
+        exposure_mjd_mid=t_mid.mjd,
+        exposure_duration=1,
+        healpixel=1,
+        observations=[],
+    )
+    assert FrameDB._compute_year_month_str(frame) == "2007-02"
+
+
+def test_add_dataset_creates_if_missing(frame_db):
```
akoumjian, line 79: Beyond scope, but since this is supposed to be idempotent and we support multiprocessing it might be good to test that the upsert / locking functionality is working.

akoumjian (approved):
Definitely headed in the right direction. Love the extra test coverage too.

moeyensj (approved):
Don't have anything to add that Alec hasn't already mentioned. Looks good!

precovery/frame_db.py
spenczar: It's a little weird, yeah. I kept it here because its purpose is as a partitioning key for the FrameDB; it's not really related to the `SourceFrame` class directly, which is just a container representing input Frame data.

It's true that `sourcecatalog` is only used when loading data into FrameDB - hence the "source" prefix. I could imagine CLI tools, for example to verify a source data file format or count the number of frames it holds, which would want to use `sourcecatalog` directly.

I'm open to reorganization if you like but making `frame_db` into a whole package with `sourcecatalog` as a module seems unnecessary to me.

precovery/frame_db.py
spenczar: Ah, so you're suggesting we drop the `add_dataset` call from in here? That does make the function really small, which is pretty nice. 

Does `load_csv` have any callers? I don't know what kind of scripts and automation we have that might be depending on its present signature.

tests/test_frame_db.py
spenczar: Multiprocessing tests are something I have not done much. I agree it's worth doing.

PR #61, opened 2023-01-25
Fix up type hints

spenczar:
This resolves all current mypy errors.

The easiest way to review this is almost certainly to go commit-by-commit.

This is based on the code in #59 rather than master, so this should wait for that to be merged.

### DIFF
--- a/precovery/frame_db.py
+++ b/precovery/frame_db.py
@@ -748,7 +748,7 @@ class FrameDB:
             f = self._current_data_file(dataset_id, year_month_str)
 
         if hasattr(observations, "__len__"):
-            logger.info(f"Writing {len(observations)} observations to {f.name}")
+            logger.info(f"Writing {len(observations)} observations to {f.name}")  # type: ignore
         else:
             logger.info(f"Writing stream of observations to {f.name}")
 
--- a/precovery/frame_db.py
+++ b/precovery/frame_db.py
@@ -132,11 +132,15 @@ class FrameIndex:
         # if it doesn't throw a warning with the command to create it
         con = engine.connect()
         curs = con.execute(
-            "SELECT name FROM sqlite_master WHERE type='table' AND name='frames';"
+            sq.text(
+                "SELECT name FROM sqlite_master WHERE type='table' AND name='frames';"
+            )
         )
         table_names = [row[0] for row in curs.fetchall()]
         if "frames" in table_names:
-            curs = con.execute("SELECT name FROM sqlite_master WHERE type = 'index';")
+            curs = con.execute(
+                sq.text("SELECT name FROM sqlite_master WHERE type = 'index';")
+            )
             index_names = [row[0] for row in curs.fetchall()]
             if "fast_query" not in index_names:
                 warning = (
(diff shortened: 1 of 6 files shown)


precovery/healpix_geom.py
```diff
@@ -1,9 +1,21 @@
+from typing import overload
+
 import healpy as hp
 import numpy as np
 import numpy.typing as npt
 
 
+@overload
```
moeyensj, line 8: Do we use the scalar definition of this function fairly often? I'm wondering if we can drop it in favor of just the array-typed one. If it is used, and I suspect that it is when HealpixFrames are initialized then no worries.

precovery/frame_db.py
```diff
@@ -747,7 +747,11 @@ def store_observations(
             self.new_data_file(dataset_id, year_month_str)
             f = self._current_data_file(dataset_id, year_month_str)
 
-        logger.info(f"Writing {len(observations)} observations to {f.name}")
+        if hasattr(observations, "__len__"):
```
moeyensj, line 754: Nice!

precovery/frame_db.py
```diff
@@ -781,19 +785,22 @@ def new_data_file(self, dataset_id: str, year_month_str: str):
         self.data_files[self._current_data_file_name(dataset_id, year_month_str)] = f
 
     def defragment(self, new_index: FrameIndex, new_db: "FrameDB"):
```
moeyensj, line 791: The defrag code definitely has been neglected over the last few months. We've been fairly consistent in making our input observations sorted by exposure_id and mjd_utc which minimizes the impact of a defrag, but I think this code is going to become more important for the quirkier datasets.

moeyensj (approved):
No major comments, feel free to merge when the other PR is merged. 

(Looks like we have some new CI errors)

precovery/frame_db.py
```diff
@@ -820,12 +828,13 @@ def defragment(self, new_index: FrameIndex, new_db: "FrameDB"):
                 )
                 new_frame = HealpixFrame(
                     id=None,
-                    obscode=cur_key[0],
-                    filter=cur_key[1],
-                    exposure_mjd_start=cur_key[2],
-                    exposure_mjd_mid=cur_key[3],
-                    exposure_duration=cur_key[4],
-                    healpixel=cur_key[5],
+                    dataset_id=cur_key[0],
```
akoumjian, line 835: Would it be too memory intense or slow things down to swap out this cursor tuple for a NamedTuple or a dataclass? Really minor thing but passing around indices gets so finicky. I know it was like this before but figured if we're updating it...

akoumjian (approved):

precovery/frame_db.py
spenczar: That seems like a good idea to me, yes. I think any performance impact should be totally negligible and it would be a lot more readable.

precovery/healpix_geom.py
spenczar: Yes, I discovered this actually only because mypy informed me, which is pretty cool. This callsite uses the scalar version: https://github.com/B612-Asteroid-Institute/precovery/blob/8defa296ded7f767fca0722348117fdcf542f15d/precovery/sourcecatalog.py#L66

precovery/frame_db.py
spenczar: It turns out this function was pretty dramatically wrong, since it depended on sorting in another query, which also wasn't updated. I'm making a separate PR to fix that.

PR #65, opened 2023-01-30
HACK: PR to trigger CI to work out test failure

spenczar:
A test seems to fail in CI that doesn't fail locally. This PR is just to crank up log levels and investigate.


spenczar:
Figured it out. The test failures have to do with the release of a new SQLAlchemy major version.

PR #66, opened 2023-01-30
Upgrade to sqlalchemy 2.0.0

moeyensj:
Mob debugging session with @spenczar and @ntellis.


spenczar (approved):
Looks good to me. Thanks for the debugging session!

PR #67, opened 2023-01-31
Use SQLAlchemy 1.x

spenczar:
Dagster is incompatible with SQLAlchemy 2, and we use dagster to run precovery jobs. Unfortunately, this means we need to revert the changes in #66.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -36,5 +36,5 @@ repos:
   - id: mypy
     additional_dependencies:
       - numpy
-      - sqlalchemy
+      - sqlalchemy<2
       - types-requests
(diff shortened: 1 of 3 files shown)


precovery/ingest/index_observations.py
```diff
@@ -109,7 +108,6 @@ def index(
         completed += 1
         logger.info(f"{completed}/{len(files)} files completed.")
     pool.close()
-    pool.join()
```
akoumjian: we can keep this line I think

akoumjian (approved):

precovery/ingest/index_observations.py
spenczar: Yes, you're right.

moeyensj (approved):

PR #68, opened 2023-01-31
Rewrite sourcecatalog ingestion for flexibility

spenczar:
Deleted precovery.sourcecatalog.SourceExposure. Its functionality was purely internal to the sourcecatalog module, and it can be accomplished using a dictionary instead, which is less code weight.

Rewrote most of sourcecatalog. Instead of primarily operating on CSV files, it breaks that step into two phases. One phase is loading in a stream of SourceObservations. This could be from a CSV file, but could also be from a DataFrame, for example. The second phase is organizing the SourceObservations into chunks grouped by HEALPixel ("SourceFrames"), suitable for loading into the database. Those two phases are now contained in separate functions.

### DIFF
--- a/precovery/frame_db.py
+++ b/precovery/frame_db.py
@@ -669,7 +669,7 @@ class FrameDB:
         limit : int, optional
             Maximum number of frames to load from the file. None means no limit.
         """
-        frames = sourcecatalog.iterate_frames(
+        frames = sourcecatalog.frames_from_csv_file(
             csv_file,
             limit,
             nside=self.healpix_nside,
--- a/precovery/frame_db.py
+++ b/precovery/frame_db.py
@@ -505,9 +505,7 @@ class FrameIndex:
             sq.Column("data_offset", sq.Integer),
             sq.Column("data_length", sq.Integer),
             # Create index on midpoint mjd, healpixel, obscode
-            sq.Index(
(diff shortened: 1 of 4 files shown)


akoumjian (approved):
I'll write up an implementation for datafram in etl repo and if we like it we can port it here afterwards

precovery/frame_db.py
```diff
@@ -505,9 +505,7 @@ def initialize_tables(self):
             sq.Column("data_offset", sq.Integer),
             sq.Column("data_length", sq.Integer),
             # Create index on midpoint mjd, healpixel, obscode
-            sq.Index(
-                "fast_query", "exposure_mjd_mid", "healpixel", "obscode", unique=True
-            ),
+            sq.Index("fast_query", "exposure_mjd_mid", "healpixel", "obscode"),
```
spenczar, line 508: I'd like @moeyensj's eyes on this - do you foresee a big impact from removing `unique=True` from this index?

moeyensj: I honestly don't have a good sense of the performance impact this change will have but it seems like a necessary one to enable the upcoming pipeline. The uniqueness constraint was originally added because it was just another test to check if the input HDF5 files were sorted correctly and that exposures were correctly grouped. We assumed up until that point that we would always have the files sorted.

moeyensj (approved):

PR #69, opened 2023-02-10
Add test case which shows failure to combine frames together

spenczar:
The `test_add_frames_dedupes_sorted_exposures` test case fails. `len(frames)` is 6 instead of 3.

### DIFF
--- a/precovery/sourcecatalog.py
+++ b/precovery/sourcecatalog.py
@@ -89,6 +89,7 @@ def bundle_into_frames(
 
             # Reset observations dictionary
             observations_by_healpixel.clear()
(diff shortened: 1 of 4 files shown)


precovery/sourcecatalog.py
```diff
@@ -89,6 +89,7 @@ def bundle_into_frames(
 
             # Reset observations dictionary
             observations_by_healpixel.clear()
+            cur_exposure_id = obs.exposure_id
```
akoumjian, line 92: aha!

akoumjian (approved):
:shipit:

Issue #70, opened 2023-02-15
Expose  allow_version_mismatch in `precovery.main.precovery`


yyang35:
I was trying to run precovery search using the precovery function in `precovery.main`, however the following error appeared. I think it would be useful to expose the `allow_version_mismatch` argument in this function. 
```python
Exception: Version mismatch: 
Running version: 0.2.dev112+g40f8bf5.d20230215
Database version: 0.2.dev108+gdbdc8b0
Use allow_version_mismatch=True to ignore this error.
```
I am working with @moeyensj on skyMapper.

PR #71, opened 2023-02-21
Allow mismatch override in main entrypoint

akoumjian:


moeyensj:
This closes #70. Thanks for the quick fix!

PR #73, opened 2023-02-28
Allow ra_sigma, dec_sigma, mag_sigma to be NaN (optional)

moeyensj:
We've run into this issue while working on a variety of datasets. Most recently, while working with @ameliaWhitlow on the ITF we were unable to index that data locally because the ITF doesn't provide any uncertainties whatsoever. Since the ITF spans observations from many different sources, assuming a fiducial value for these uncertainties is not the best approach.


moeyensj:
TIL that footnotes now also render on GH markdown.

tests/test_frame_db.py
```diff
@@ -250,6 +251,62 @@ def test_add_frames(frame_db):
     assert len(stored_obs1) == 2
 
 
+def test_add_frames_nullable(frame_db):
+    """Add a frame with an observation that contains nullable quantities (represented as NaN)
+    and insure that they are stored and retrieved correctly.
```
spenczar:
```suggestion
    and ensure that they are stored and retrieved correctly.
```

README.md
```diff
@@ -109,29 +109,27 @@ efficient format. Input files should be sorted by ascending time.
...
+| obs_id | None | str | Unique observation ID for the observation |
 | exposure_id | None | str | Exposure or Image ID from which observation was measured |
-| mjd | days | float | MJD of the observation in UTC** |
+| mjd | days | float | MJD of the observation in UTC[^1] |
 | ra  | degree | float | Right Ascension (J2000) |
 | dec  | degree | float | Declination (J2000) |
-| ra_sigma  | degree | float | 1-sigma uncertainty in Right Ascension |
-| dec_sigma  | degree | float | 1-sigma uncertainty in Declination |
+| ra_sigma  | degree | float | 1-sigma uncertainty in Right Ascension (Nullable)[^2] |
+| dec_sigma  | degree | float | 1-sigma uncertainty in Declination (Nullable)[^2] |
```
spenczar: I don't think "Nullable" is quite an accurate description. They're still always floats, they aren't ever valued as `None`. It's more than NaN is explicitly supposed to be expected.

README.md
```diff
@@ -109,29 +109,27 @@ efficient format. Input files should be sorted by ascending time.
...
 
 ## Precovery Results
 
 |Name|Unit|Type|Description|
 |---|---|---|---|
-| mjd | days | float | MJD of the observation in UTC** |
+| mjd | days | float | MJD of the observation in UTC[^1] |
 | ra_deg  | degree | float | Right Ascension (J2000) |
 | dec_deg  | degree | float | Declination (J2000) |
 | ra_sigma_arcsec  | arcsecond | float | 1-sigma uncertainty in Right Ascension |
```
spenczar: This is really the area that's more important for documentation: Precovery's *outputs* might be NaNs for certain fields.

spenczar (approved):
This looks good.

I think it's more important to document the changes on the *results* side than the inputs side. This is sort of an application of [Postel's Law](https://en.wikipedia.org/wiki/Robustness_principle): systems should be permissive in what they accept, and strict in what they provide. We don't _want_ NaNs on the input, but we'll accept them and handle them. But we need to be extremely precise about the boundaries of the outputs of our system. 

This is a good rule for all API design in software, really. Nobody is going to be upset if you accept too much (if your `sqrt(float: value) -> float` function also accepts `int`s, great!), but they _will_ be upset if it provides too much (if your `sqrt(float: value) -> float` sometimes returns None, they have every reason to be annoyed).

README.md
moeyensj: How about swapping to "Optional"? I can update the footnote to say that the column should still be included in the CSV but it should be empty. 

(We could relax this too and make the presence of the column actually optional.)

README.md
moeyensj: I'll update the description and describe frame candidates as well.

README.md
spenczar: Yeah, I think "optional" is a better word for it.

tests/test_frame_db.py
moeyensj: Done.

PR #77, opened 2023-03-07
Sort candidates by MJD and observation ID

moeyensj:
Sort candidates by MJD, and then by Observation ID. If sorting a FrameCandidate, it will be sort on midpoint exposure time with a dummy observation ID. As precovery currently works, one orbit will never return the same frame more than once.

### DIFF
--- a/precovery/precovery_db.py
+++ b/precovery/precovery_db.py
@@ -145,7 +145,6 @@ class PrecoveryDatabase:
         self,
         orbit: Orbit,
         tolerance: float = 30 * ARCSEC,
-        max_matches: Optional[int] = None,
         start_mjd: Optional[float] = None,
         end_mjd: Optional[float] = None,
         window_size: int = 7,
@@ -186,7 +185,6 @@ class PrecoveryDatabase:
             if end_mjd is None:
                 end_mjd = last
 
-        n = 0
         logger.info(
             "precovering orbit %s from %.5f to %.5f, window=%d",
             orbit.orbit_id,
@@ -214,9 +212,6 @@ class PrecoveryDatabase:
             )
             for result in matches:
                 yield result
-                n += 1
-                if max_matches is not None and n >= max_matches:
-                    return
 
     def _check_windows(
         self,
--- a/precovery/precovery_db.py
+++ b/precovery/precovery_db.py
@@ -83,16 +83,26 @@ def sort_candidates(candidates: List[Union[PrecoveryCandidate, FrameCandidate]])
         Sorted list of candidates.
     """
     mjds = []
+    observation_ids = []
     for candidate_i in candidates:
         if isinstance(candidate_i, PrecoveryCandidate):
             mjds.append(candidate_i.mjd)
+            observation_ids.append(candidate_i.observation_id)
         elif isinstance(candidate_i, FrameCandidate):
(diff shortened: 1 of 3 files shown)


precovery/precovery_db.py
```diff
@@ -67,6 +68,43 @@ class FrameCandidate:
...
+    mjds = []
+    observation_ids = []
+    for candidate_i in candidates:
+        if isinstance(candidate_i, PrecoveryCandidate):
+            mjds.append(candidate_i.mjd)
+            observation_ids.append(candidate_i.observation_id)
+        elif isinstance(candidate_i, FrameCandidate):
+            mjds.append(candidate_i.exposure_mjd_mid)
+            # Add empty observation ID since frame candidates don't have one
+            # Frames, so this is fine
```
spenczar: This comment looks a little garbled

precovery/precovery_db.py
```diff
@@ -67,6 +68,43 @@ class FrameCandidate:
...
+            observation_ids.append("")
+        else:
+            raise TypeError("Candidates must be PrecoveryCandidate or FrameCandidate")
+
+    cand_array = np.array(candidates)
+    df = pd.DataFrame({"mjd": mjds, "observation_id": observation_ids})
+    df.sort_values(by=["mjd", "observation_id"], inplace=True)
+    return cand_array[df.index.values].tolist()
```
spenczar: Of course, what you have is fine, and looks correct. But I'd just like to share the stdlib, non-pandas way of doing this, which is in one line:

```suggestion
    return sorted(candidates, lambda c: (c.mjd, c.observation_id) if instance(c, PrecoveryCandidate) else (c.exposure_mjd_mid, ""))
```

spenczar: If the one-liner is too crazy, you could do

```suggestion
    def sort_key(candidate):
        if isinstance(candidate, PrecoveryCandidate):
            return (candidate.mjd, candidate.observation_id)
        elif isinstance(candidate, FrameCandidate):
            return (candidate.exposure_mjd_mid, "")
        else:
            raise TypeError(f"Candidates must be PrecoveryCandidate or FrameCandidate, but got {type(candidate)}")

    return sorted(candidates, key=sort_key)
```

spenczar (commented):
Could you also update https://github.com/B612-Asteroid-Institute/precovery/blob/main/precovery/main.py#L56? I think `precovery.main.precover_many` was added after you branched off of main, so you may need to rebase.

precovery/precovery_db.py
moeyensj: Your solution is faster than pandas (silly in-function dataframe creation \*shakes fists\*)!

See: 
```python
%%timeit 
sort_candidates(cands)
```
`700 µs ± 3.68 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)`

```python
%%timeit 
sort_candidates_no_pandas(cands)
```
`770 ns ± 4.21 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)`

spenczar: Wow, a factor of 1000 faster too! This is truly a great day for the anti-pandas resistance

spenczar (approved):

PR #78, opened 2023-03-10
Fix get_frames_for_ra_dec 

moeyensj:


moeyensj:
Happy to drop the last commit if I misunderstood the intention. I removed the nside parameter from the `get_frames_for_ra_dec` function signature since the FrameDB nside must equal the search nside for this function to work. Once we get Tanay's downsampling code we may want to revisit this.

spenczar (approved):
Yikes, thanks.

PR #80, opened 2023-03-14
Delete unused FrameBundleDescription and FrameWindow classes and methods

spenczar:
These aren't actually used anywhere. They were an abstraction which was removed in 0315822b2295dcccca18d714c0e02285a4bcac2a, in favor of the windows-of-n-days approach.


moeyensj (approved):
Only removed lines and unit tests pass. 'Tis a good day.

PR #82, opened 2023-03-14
Add ObservationArray class to clarify distance calculations

spenczar:
Added a new file, precovery/observation.py, to house the existing Observation class, and a new ObservationArray class, which is currently unused.

Added a method on Ephemeris to compute its distance from an ObservationArray.

The payoff to this PR is really in #83. I separated the two phases to make review a little cleaner.

### DIFF
--- a/precovery/orbit.py
+++ b/precovery/orbit.py
@@ -7,6 +7,9 @@ import numpy.typing as npt
 import requests as req
 from astropy.time import Time
 
+from .observation import ObservationArray
+from .spherical_geom import haversine_distance_deg
+
 try:
     import pyoorb
 except ImportError:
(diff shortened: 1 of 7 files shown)


precovery/orbit.py
```diff
@@ -381,15 +387,27 @@ def precover_remote(
 
 
 class Ephemeris:
-    def __init__(self, raw_data: npt.NDArray[np.float64]):
-        self._raw_data = raw_data
+    def __init__(self, mjd, ra, dec, ra_velocity, dec_velocity):
```
akoumjian, line 390: worth adding typing here, or not yet?

akoumjian (approved):
Looks good. What are your thoughts on switching everything over to keyword arguments when we aren't using data classes? I always get uncomfortable initializing classes with > 4 arguments, all of which are floats or strings and relying on positional arguments being sent in correctly.

precovery/orbit.py
akoumjian: ..aand just saw it on the other PR

PR #83, opened 2023-03-14
refactor: Rewrite inner loop of precover for clarity, and use N-body propagation for per-obs timestamped datasets

spenczar:
This completes the refactoring started in #82. The inner loop is broken into many smaller functions, constructors are added for candidates. All the `zip` stuff is replaced with structured arrays and constructors which pull values out of objects.

### DIFF
--- a/precovery/orbit.py
+++ b/precovery/orbit.py
@@ -387,7 +387,9 @@ class Orbit:
 
 
 class Ephemeris:
-    def __init__(self, mjd, ra, dec, ra_velocity, dec_velocity):
+    def __init__(
+        self, mjd: float, ra: float, dec: float, ra_velocity: float, dec_velocity: float
+    ):
         self.mjd = mjd
         self.ra = ra
         self.dec = dec
--- a/precovery/precovery_db.py
+++ b/precovery/precovery_db.py
@@ -7,10 +7,10 @@ from typing import Iterable, Iterator, List, Optional, Union
 import numpy as np
 
 from .config import Config, DefaultConfig
-from .frame_db import FrameDB, FrameIndex
+from .frame_db import FrameDB, FrameIndex, HealpixFrame
 from .healpix_geom import radec_to_healpixel
-from .observation import Observation
-from .orbit import EpochTimescale, Orbit, PropagationIntegrator
+from .observation import Observation, ObservationArray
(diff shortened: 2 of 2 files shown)


precovery/orbit.py
```diff
@@ -387,7 +387,9 @@ def precover_remote(
 
 
 class Ephemeris:
-    def __init__(self, mjd, ra, dec, ra_velocity, dec_velocity):
+    def __init__(
+        self, mjd: float, ra: float, dec: float, ra_velocity: float, dec_velocity: float
```
akoumjian, line 391: Well look at that! There's the typing :smile:

precovery/precovery_db.py
```diff
@@ -51,6 +51,51 @@ class PrecoveryCandidate:
...
+        # nside value
+        healpix_id = int(radec_to_healpixel(ephem.ra, ephem.dec, nside=CANDIDATE_NSIDE))
+
+        return cls(
+            # Observation data:
+            observation_id=obs.id.decode(),
+            mjd=obs.mjd,
+            ra_deg=obs.ra,
+            dec_deg=obs.dec,
+            ra_sigma_arcsec=obs.ra_sigma / ARCSEC,
```
akoumjian, line 74: How come we have to convert the sigmas here but not ra/dec itself?

akoumjian (approved):
Nice changes

precovery/precovery_db.py
spenczar: I don't really know. They're both stored internally in degrees, but we present error estimates in arcseconds. This was introduced in https://github.com/B612-Asteroid-Institute/precovery/commit/4c603b3732e9630cce30d3f391ab0e1ef9d40400. Maybe @moeyensj can explain the reasoning.

moeyensj: @spenczar @akoumjian This is purely a conversion to the most easily interpretable scale for orbital uncertainty mapped to observations. Actual on-sky positions remain in degrees while all uncertainties go to arcseconds because errors are roughly on that scale. So we set tolerances in arcseconds and report astrometric errors to the MPC in arcseconds. One fun tidbit is that if we wanted to calculate the chi^2 per coordinate with our current structure we would just do (residual_ra_arcsec)^2 / (ra_sigma_arcsec)^2 as well.

PR #85, opened 2023-03-22
Proposal: Switch to non-dataframe tupled results

akoumjian:
This PR switches the primary interface for `precovery.precover` from returning a `pd.DataFrame` of unioned precovery candidates and frame candidates into two separate return values. The new format takes the form `Tuple[List[PrecoveryCandidate, FrameCandidate]]`. The second major change eliminates the `include_frame_candidates` option, opting instead to always return frame candidates.

Old

```python
>>> df = precovery.precover(...)
>>> df.columns
Index(['mjd', 'ra_deg', 'dec_deg', 'ra_sigma_arcsec', 'dec_sigma_arcsec',
       'mag', 'mag_sigma', 'filter', 'obscode', 'exposure_id',
       'exposure_mjd_start', 'exposure_mjd_mid', 'exposure_duration',
       'observation_id', 'healpix_id', 'pred_ra_deg', 'pred_dec_deg',
       'pred_vra_degpday', 'pred_vdec_degpday', 'delta_ra_arcsec',
       'delta_dec_arcsec', 'distance_arcsec', 'dataset_id', 'orbit_id'],
      dtype='object')
```

New

```python
>>> precovery_candidates, frame_candidates = precovery.precover()
>>> print(precovery_candidates)
[PrecoveryCandidate(mjd=50000.0, ra_deg=205.88903218124665, dec_deg=-15.381163692732548, ra_sigma_arcsec=10800.0, dec_sigma_arcsec=14400.0, mag=5.0, mag_sigma=6.0, exposure_mjd_start=50000.0, exposure_mjd_mid=50000.0, filter='filter', obscode='I41', exposure_id='exposure', exposure_duration=30.0, observation_id='bjewnltujtgbffge', healpix_id=6715579247, pred_ra_deg=205.88903218124662, pred_dec_deg=-15.381163692732573, pred_vra_degpday=1.5288441441416185, pred_vdec_degpday=-0.8115017070879296, delta_ra_arcsec=1.0231815394945443e-10, delta_dec_arcsec=8.952838470577262e-11, distance_arcsec=1.9897641567037682e-10, dataset_id='test_dataset_2'), ...]
```

## "why two values instead of a single value?" 
The two objects being returned have distinct schemas and represent distinct types. When performing work on the results of a precovery search, most downstream tasks involve only the `PrecoveryCandidate` matches. Returning the results as a unioned iterable requires every downstream consumer to filter them out with type checking. Even the cases where both matches and misses are wanted, typically the sets are separated when additional work is performed.

A unioned dataframe is doubly difficult because now the filtering must be done heuristically by analyzing column values (e.g. `df.dropna(subset=["mjd"])`) and hoping that the resulting subset is valid. This unioned schema also causes problems when attempting to cast column types, but some fields are null, such as casting `observation_id` to `str` which in turn causes the frame candidate row `NaNs` to become literal `"nan"`. 

## Why not two dataframes?
Dataframes have very limited typing semantics. It is difficult to make guarantees about nullability. By returning more specific types, downstream consumers can more easily inspect and react to individual rows. I've added two utility functions to initialize dataframes from the dataclasses when wanted (`as_df = candidates_to_dataframes(precovery_candidates)`).

## Why always return frame candidates?
There is almost no perceived performance benefit in avoiding returning the frame candidates. By always returning a tuple (matches, misses), the contents of the return values are never surprising. If a downstream consumer does not care about the frame candidates, it can simply discard them. 

## Other changes
I moved the sorting logic further up into the main `PrecoveryDatabase.precover` function (`precovery.precovery_db.sift_candidates`). This ensures that the `precovery.main.precover` function is just syntactic sugar for the `db.precover` function and that the two return identical results. Previously the functional version performed separate sorting and type casting (observation_id -> str) that differed from the streaming style results of the db method.

### DIFF
--- a/precovery/utils.py
+++ b/precovery/utils.py
@@ -1,0 +1,21 @@
+from typing import List
+
+import pandas as pd
+
+from .precovery_db import FrameCandidate, PrecoveryCandidate
+
+
+def candidates_to_dataframe(candidates: List[PrecoveryCandidate]):
+    """Convert a list of precovery candidates to a pandas dataframe."""
+    if len(candidates) == 0:
+        return pd.DataFrame()
+    else:
+        return pd.DataFrame([c.to_dict() for c in candidates])
+
+
+def frames_to_dataframe(frames: List[FrameCandidate]):
+    """Convert a list of frame candidates to a pandas dataframe."""
+    if len(frames) == 0:
+        return pd.DataFrame()
+    else:
+        return pd.DataFrame([c.to_dict() for c in frames])
--- a/tests/test_precovery_index.py
+++ b/tests/test_precovery_index.py
@@ -9,6 +9,7 @@ import pytest
 from precovery.ingest import index
 from precovery.main import precover
 from precovery.orbit import EpochTimescale, Orbit
+from precovery.utils import candidates_to_dataframe
 
 from .testutils import requires_openorb_data
 
@@ -90,16 +91,20 @@ def test_precovery(test_db_dir):
 
     # For each sample orbit, validate we get all the observations we planted
     for orbit in orbits_keplerian:
-        results = precover(orbit, test_db_dir, tolerance=1 / 3600, window_size=1)
+        matches, misses = precover(
+            orbit, test_db_dir, tolerance=1 / 3600, window_size=1
+        )
 
         orbit_name = orbit_name_mapping[orbit.orbit_id]
         object_observations = observations_df[
             observations_df["object_id"] == orbit_name
         ]
-        assert len(results) == len(object_observations)
-        assert len(results) > 0
+        assert len(matches) == len(object_observations)
+        assert len(matches) > 0
 
-        results.rename(
+        matches = candidates_to_dataframe(matches)
+
+        matches.rename(
             columns={
                 "ra_deg": "ra",
                 "dec_deg": "dec",
@@ -111,8 +116,8 @@ def test_precovery(test_db_dir):
             inplace=True,
         )
 
-        results["ra_sigma"] /= 3600.0
-        results["dec_sigma"] /= 3600.0
+        matches["ra_sigma"] /= 3600.0
+        matches["dec_sigma"] /= 3600.0
 
         # We are assuming that both the test observation file and the results
         # are sorted by mjd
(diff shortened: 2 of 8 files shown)


tests/test_precovery_index.py
```diff
@@ -90,16 +91,20 @@ def test_precovery(test_db_dir):
 
     # For each sample orbit, validate we get all the observations we planted
     for orbit in orbits_keplerian:
-        results = precover(orbit, test_db_dir, tolerance=1 / 3600, window_size=1)
+        matches, misses = precover(
+            orbit, test_db_dir, tolerance=1 / 3600, window_size=1
+        )
```
spenczar, line 96: I like this language - "matches and misses" - a lot more than "PrecoveryCandidates and FrameCandidates." Maybe we can switch to that?

spenczar (approved):
I love it

tests/test_precovery_index.py
akoumjian: I was thinking about that too. If the rest of the team is okay with these high level changes I'd be happy to use that new language. In particular, FrameCandidate had me confused for a long time initially.

precovery/utils.py
```diff
@@ -0,0 +1,21 @@
+from typing import List
+
+import pandas as pd
+
+from .precovery_db import FrameCandidate, PrecoveryCandidate
+
+
+def candidates_to_dataframe(candidates: List[PrecoveryCandidate]):
```
moeyensj, line 8: Thanks for including these!

tests/test_precovery_index.py
moeyensj: I like this too. Though I'm wondering if we can maintain some clarity on the fact that these really aren't "matches" without additional validation (OD, cutouts, etc..). I'll defer to the two of you to decide if you think that matters.

moeyensj (approved):

tests/test_precovery_index.py
akoumjian: That's an interesting point. We don't want anyone naively assuming that if precovery returns a result, it is a match for their object. However, it's also true that what is being returned is a match for the parameters sent in.

PR #89, opened 2023-12-07
adam_core interfaces

ntellis:
This PR adds:

- to/from `adam_core` functions for `precovery.Orbit`
- quivr tables that mimic `PrecoveryCandidate` and `FrameCandidate`, with constructors from a list of either. This table is flat, but includes convenience methods to construct the `adam_core` classes (ex `Residuals`, `Ephemeris` for the predicted positions)

As written, `precover()` and `precover_many()` both accept either a `precovery.Orbit` or an `adam_core` `Orbits`. They _output_ the old `PrecoveryCandidate` and `FrameCandidate`, though - this way they don't break anything existing in cutouts or adam_etl.

To get the more useful `PrecoveryCandidateQv`, the constructor is:

```
PrecoveryCandidatesQv.from_precovery_candidates([cand1, cand2, cand3])
```

I think it will be easiest to just call this after running precovery normally, where desired.


ntellis:
I also borrowed `drop_duplicates` from thor. This could probably be promoted to quivr

ntellis:
Dependencies are borked here, may need to update to using pyproject.toml, as well as change some things with pyoorb in order to get this working.

precovery/candidates_qv.py
```diff
@@ -0,0 +1,298 @@
...
+        .column(f"index_{agg_func}")
+    )
+
+    # Take the indices from the flattened table and use them to index into the original table
+    return table.take(indices)
+
+
+class PrecoveryCandidatesQv(qv.Table):
+    # copy all the fields from PrecoveryCandidate
+    mjd = Timestamp.as_column()
```
ntellis: Change to Time

precovery/candidates_qv.py
```diff
@@ -0,0 +1,298 @@
...
+                                sig_lat.as_py(),
+                                np.nan,
+                                np.nan,
+                                np.nan,
+                            ],
+                            np.float64,
+                        )
+                        for sig_lon, sig_lat in zip(
+                            self.ra_sigma_arcsec,
+                            self.dec_sigma_arcsec,
```
ntellis: This could be vectorized

precovery/candidates_qv.py
```diff
@@ -0,0 +1,298 @@
...
+        flattened_table.group_by(subset, use_threads=False)
+        .aggregate([("index", agg_func)])
+        .column(f"index_{agg_func}")
+    )
+
+    # Take the indices from the flattened table and use them to index into the original table
+    return table.take(indices)
+
+
+class PrecoveryCandidatesQv(qv.Table):
```
ntellis: This wants a to_dataclass

ntellis:
Adopt dockerfile structure for adam_core/adam_etl for installing pyoorb, switch to pyproject.toml due to recent setuptools main version.

tests/test_precovery_index.py
```diff
@@ -95,6 +95,10 @@ def test_precovery(test_db_dir):
             orbit, test_db_dir, tolerance=1 / 3600, window_size=1
         )
 
+        # convert back to dataclasses
+        matches = matches.to_dataclass()
+        misses = misses.to_frame_candidates()
```
akoumjian: These method names should follow the same pattern.

precovery/precovery_db.py
```diff
@@ -183,6 +193,287 @@ def sift_candidates(
     return precovery_candidates, frame_candidates
 
 
+class PrecoveryCandidatesQv(qv.Table):
```
akoumjian, line 196: I think we should just call this `PrecoveryCandidates` and call the old ones `PrecoveryCandidateDataclass`, or some such. Only because we want to move the quivr tables to be the defaults over the long run.

akoumjian (approved):

PR #99, opened 2024-09-18
Add parallel precovery_db implementation that uses adam_core orbits/propagation

ntellis:
This adds a new class, `PrecoveryDatabaseQv` which uses `adam_core` propagation and Orbits / ephemerides. 

Currently on first run in a new environment, precovery is pretty slow. This appears to be due to jax, where compile time is expensive for the @jit-ed `propagate_2body` in adam core. My understanding of this is limited, but there is some indication that this is also related to the input size. Just based on the runtimes profiled, it seems that recompilation needs to occur to propagate to 500 frames, even if the function has been compiled to propagate to say, 300. For this reason, precovery over the whole of NSC can take 5 minutes for orbit A, then 1 minute if run a second time, but it may still take 5 minutes to run for orbit B, all in the same environment. This is the main bottleneck right now, and hamstrings precovery pretty significantly if it's run in a fresh environment (like a k8s pod, for example).

The other current expense is in creating the observers class repeatedly via `Observers.from_code`, which involves hitting the spice files. This could be sped up.

Also included in this PR are the transition to PDM, removal of pyoorb as a dependency.


ntellis:
The rest of precovery has now been migrated to use the adam_core `Orbits` class and the `Precovery/FrameCandidates` quivr tables. Was able to remove a lot of now unused code, including the entire old Orbit class, the precovery candidate and frame candidate dataclasses, the conversions from the same to the quivr types, etc.

The test dataset (in `/tests/data/index/`) has been regenerated with Assist ephemeris generation, get_orbits.py now generates `/tests/data/sample_orbits.parquet`, which is what the tests read from.

There is now a new test decorator that checks for the existence of `ASSIST_DATA_DIR` in the env - this is the location of the jpl ephemeris files. We skip index, precovery search tests without it. test_index.py only runs a precovery search on one orbit for now, due to how long it takes (~5 min, still mostly dominated by 2-body propagation time).

Core precovery is propagator-agnostic, but all of the tests initialize an `AssistPropagator` before doing any propagation or ephemeris generation.

Consumers will need to be updated:
- `db.precover` (and by extension main.precover, precover_many, etc) now takes an adam_core `Propagator` as an argument
- precover functions return a tuple of `(PrecoveryCandidates, FrameCandidates)`, which have been renamed from `PrecoveryCandidatesQv` and `FrameCandidatesQv`

pyproject.toml
```diff
@@ -15,22 +15,24 @@ authors = [
...
-    "Programming Language :: Python",
     "Programming Language :: Python :: 3.10",
     "Topic :: Scientific/Engineering :: Astronomy",
     "Topic :: Scientific/Engineering :: Physics",
 ]
 
 dependencies = [
-    "adam_core>=0.2.0",
+    "adam_core[assist]>=0.2.0",
+    "adam_assist",
```
akoumjian:
```suggestion
    "adam_core>=0.2.3",
```

Let's move `adam-assist>=0.1.2` to the dev optional-dependencies. Also let's require more recent adam_core.

precovery/main.py
```diff
@@ -22,7 +23,10 @@ def precover_many(
     allow_version_mismatch: bool = False,
     datasets: Optional[set[str]] = None,
     n_workers: int = multiprocessing.cpu_count(),
-) -> Tuple[PrecoveryCandidatesQv, FrameCandidatesQv]:
+    propagator: Optional[
+        Propagator
+    ] = None,  # should we initialize an assist propagator?
```
akoumjian: Not sure. We should probably just throw an error for now.

precovery/main.py
```diff
@@ -61,15 +66,15 @@ def precover_many(
...
 def precover_worker(
-    orbit: AdamOrbits,
+    orbit: Orbits,
     database_directory: str,
     tolerance: float = 1 / 3600,
     start_mjd: Optional[float] = None,
     end_mjd: Optional[float] = None,
     window_size: int = 7,
     allow_version_mismatch: bool = False,
     datasets: Optional[set[str]] = None,
```
akoumjian, line 74: We should be passing through the propagator class here and into the precover function.

precovery/main.py
```diff
@@ -91,21 +96,24 @@ def precover_worker(
...
-) -> Tuple[PrecoveryCandidatesQv, FrameCandidatesQv]:
+    propagator: Optional[
+        Propagator
+    ] = None,  # should we initialize an assist propagator?
```
akoumjian:
```suggestion
    propagator: Optional[Type[Propagator]] = None
```
We should accept a class, rather than an instance. This makes it easier to do the multiprocessing, since we want to make sure we aren't re-using instances of the propagator across threads.

precovery/precovery_db.py
```diff
@@ -605,13 +406,16 @@ def create(
...
+        orbit: Orbits,
         tolerance: float = 30 * ARCSEC,
         start_mjd: Optional[float] = None,
         end_mjd: Optional[float] = None,
         window_size: int = 7,
         datasets: Optional[set[str]] = None,
-    ) -> Tuple[List[PrecoveryCandidate], List[FrameCandidate]]:
+        propagator: Optional[
+            Propagator
+        ] = None,  # should we initialize an assist propagator?
```
akoumjian:
```suggestion
    propagator: Optional[Type[Propagator]] = None
```

Accept a propagator class. Then inside the function we initialize it.

precovery/precovery_db.py
```diff
@@ -669,74 +472,62 @@ def precover(
...
-
         # group windows by obscodes so that many windows can be searched at once
-        matches = []
-        for obscode, obs_windows in itertools.groupby(
-            windows, key=lambda pair: pair[1]
-        ):
-            mjds = [window[0] for window in obs_windows]
-            matches_window = self._check_windows(
-                mjds,
+        candidates = []
```
akoumjian, line 478:
```suggestion
        prop = Propagator()
        candidates = []
```

precovery/precovery_db.py
```diff
@@ -669,74 +472,62 @@ def precover(
...
+            logger.info("searching windows for obscode %s", obscode)
+            # create our Timestamp ghere
+            times = Timestamp.from_mjd(mjds, scale="utc")
+
+            candidates_obscode, frame_candidates_obscode = self._check_windows(
+                times,
                 obscode,
                 orbit,
                 tolerance,
+                propagator,
```
akoumjian:
```suggestion
                prop,
```

Here we send in the instance.

precovery/precovery_db.py
```diff
@@ -669,74 +472,62 @@ def precover(
...
         # convert these to our new output formats
-        return PrecoveryCandidatesQv.from_dataclass(
-            precovery_candidates, source_orbit_id=orbit_id
-        ), FrameCandidatesQv.from_dataclass(frame_candidates, source_orbit_id=orbit_id)
+        return qv.concatenate(candidates), qv.concatenate(frame_candidates)
 
     def _check_windows(
         self,
-        window_midpoints: Iterable[float],
+        window_midpoints: Timestamp,
```
akoumjian, line 504: :thumbsup:

precovery/precovery_db.py
```diff
@@ -669,74 +472,62 @@ def precover(
...
         """
         # Propagate the orbit with n-body to every window center
-        # Since the window midpoints are calculated from the observations
-        # in the database then they are in the UTC timescale so let's use that
-        orbit_propagated = orbit.propagate(
-            window_midpoints,
-            PropagationIntegrator.N_BODY,
-            time_scale=EpochTimescale.UTC,
+        orbit_propagated = propagator.propagate_orbits(orbit, window_midpoints)
+        cov = CoordinateCovariances.from_kwargs(
```
akoumjian: I can't remember, did this end up mattering at all?

precovery/precovery_db.py
```diff
@@ -791,81 +587,102 @@ def _check_window(
...
+        ephemeris = propagator.generate_ephemeris(orbit, observers)
+        frames = []
+        for mjd, healpixel in zip(mjds, healpixels):
+            frames.append(self.frames.idx.get_frames(obscode, mjd, healpixel, datasets))
+        # logger.debug(
+        #     "checking frames for healpix=%d obscode=%s mjd=%f",
+        #     healpixel,
+        #     obscode,
+        #     mjd,
+        # )
```
akoumjian:
```suggestion
```

precovery/utils.py
```diff
@@ -1,21 +1,56 @@
...
+
+    agg_func = keep
+    indices = (
+        flattened_table.group_by(subset, use_threads=False)
+        .aggregate([("index", agg_func)])
+        .column(f"index_{agg_func}")
+    )
+
+    # Take the indices from the flattened table and use them to index into the original table
+    return table.take(indices)
```
akoumjian: FYI I think this got merged into quivr and is available in 0.7.4a2

akoumjian: FYI I think this got merged into quivr and is available in 0.7.4a2

akoumjian (changes requested):
Some suggestions.

precovery/precovery_db.py
ntellis: Ah good catch - no this is left in from when we were debugging. I'll remove it.

precovery/utils.py
ntellis: Right now adam_core pins to 0.7.3a1 - should I leave it in for now and we add that to the next adam_core release?

precovery/main.py
```diff
@@ -61,18 +64,22 @@ def precover_many(
...
     datasets: Optional[set[str]] = None,
-) -> Tuple[PrecoveryCandidatesQv, FrameCandidatesQv]:
+    propagator: Optional[Type[Propagator]] = None,
+) -> Tuple[PrecoveryCandidates, FrameCandidates]:
     """
     Wraps the precover function to return the orbit_id for mapping.
     """
+
+    # initialize our propagator
+    propagator_instance = propagator() if propagator is not None else None
```
akoumjian: We should be passing a class here too and only initializing inside the db.precover function I think. Perhaps we should change the argument name to `propagator_class` ?

akoumjian (changes requested):
Slight mismatch on class vs instance. Arguments everywhere except for `db.precover` should be class, I think, and potentially change the arguments to `propagator_class`.

ntellis:
Incorporated those suggestions @akoumjian, think this is ready to merge?

precovery/main.py
ntellis: I adopted `propagator_class` throughout

